PR #13, opened 2025-04-01
get_logs tool call

gregnr:
Allows the LLM to fetch logs for a Supabase project by service type (api, postgres, edge functions, auth, storage, realtime). Helps with debugging.

### DIFF
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -355,9 +355,7 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
                   ref: project_id,
                 },
                 query: {
-                  // Limit to the last minute
-                  iso_timestamp_start: new Date(now - 60 * 1000).toISOString(),
-                  iso_timestamp_end: new Date(now).toISOString(),
+                  // Omitting start and end time defaults to the last minute
                   sql: getLogQuery(service),
                 },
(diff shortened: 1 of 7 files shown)


packages/mcp-server-supabase/src/server.ts
```diff
@@ -328,6 +329,47 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
...
+            '/v1/projects/{ref}/analytics/endpoints/logs.all',
+            {
+              params: {
+                path: {
+                  ref: project_id,
+                },
+                query: {
+                  // Limit to the last minute
+                  iso_timestamp_start: new Date(now - 60 * 1000).toISOString(),
+                  iso_timestamp_end: new Date(now).toISOString(),
```
sweatybridge, line 360:
```suggestion
                  iso_timestamp_end: new Date(now + 5 * 60 * 1000).toISOString(),
```

Shall we set the end time to 5 minutes in the future to account for any local clock drift?

sweatybridge (approved):

packages/mcp-server-supabase/src/server.ts
sweatybridge: Or perhaps we can omit both timestamps entirely since the default is last 1 minute.

https://github.com/orgs/supabase/discussions/34634

PR #21, opened 2025-04-04
feat: prompt to clarify project cost

gregnr:
Adds prompt to `create_project` description that clarifies the costs involved when creating new projects and asks the LLM to share this with the user before creating new projects.

### DIFF
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -122,7 +122,7 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
       }),
       create_project: tool({
         description:
-          'Creates a new Supabase project. Always ask the user which organization to create the project in. The project can take a few minutes to initialize - use `getProject` to check the status.',
+          'Creates a new Supabase project. Always ask the user which organization to create the project in. Each new project can have additional costs: If on a free org, the user gets 2 projects for free. On a paid org, the user should reference https://supabase.com/pricing. Be sure the user understands this before creating new projects. The project can take a few minutes to initialize - use `getProject` to check the status.',
         parameters: z.object({
           name: z.string().describe('The name of the project'),
           region: z.optional(


sweatybridge (approved):

packages/mcp-server-supabase/src/server.ts
```diff
@@ -444,12 +508,26 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
...
-            .describe('Name of the branch to create'),
+            .describe('The cost confirmation ID. Call `confirm_cost` first.'),
         }),
-        execute: async ({ project_id, name }) => {
+        execute: async ({ project_id, name, confirm_cost_id }) => {
+          if (!confirm_cost_id) {
+            throw new Error(
+              'User must confirm understanding of costs before creating a branch.'
+            );
+          }
```
sweatybridge, line 521: Is confirmation required every time or just the first branch? Since branches can be created and deleted much more frequently than projects, it would be quite disruptive to prompt every time.

packages/mcp-server-supabase/src/server.ts
```diff
@@ -444,12 +508,26 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
           'Creates a development branch on a Supabase project. This will apply all migrations from the main project to a fresh branch database. Note that production data will not carry over. The branch will get its own project_id via the resulting project_ref. Use this ID to execute queries and migrations on the branch.',
         parameters: z.object({
           project_id: z.string(),
-          name: z
+          name: z.string().describe('Name of the branch to create'),
+          confirm_cost_id: z
             .string()
-            .default('develop')
```
sweatybridge, line 449: Any reason that we are removing the default name?

packages/mcp-server-supabase/src/pricing.ts
```diff
@@ -0,0 +1,72 @@
...
+      project.organization_id === orgId && project.status === 'ACTIVE_HEALTHY'
+  );
+
+  let amount = 0;
+
+  if (org.plan !== 'free') {
+    // If the organization is on a paid plan, the first project is included
+    if (activeProjects.length === 0) {
+      amount = 0;
+    } else {
```
sweatybridge, line 59:
```suggestion
    if (activeProjects.length > 0) {
```

nitpick

packages/mcp-server-supabase/src/server.ts
```diff
@@ -477,6 +555,7 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
               },
               body: {
                 branch_name: 'main',
+                git_branch: 'main',
```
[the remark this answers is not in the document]
gregnr: This probably should have been in a different PR, but the reason why I added this is because I noticed `git_branch` gets set to the original branch name. So after running the PATCH which updates `branch_name` to main, the `git_branch` still had the old original branch name (eg. `develop`). This seemed confusing to me but maybe it doesn't matter?

packages/mcp-server-supabase/src/server.ts
gregnr: Currently yes it is required every time. Agreed this is not ideal, though technically today the LLM can reuse the same confirmation ID for multiple creations. TBH I can't think of a good way to require this only the first time, but not future times, given how tool definitions work (eg. tool schemas are fixed, so you can't require it at one point then make it optional later). We could make the confirmation ID optional with implicit logic that throws and error if you don't provide it the first time, but this gets a bit messy.

Given our preference to err on being more explicit, maybe we keep as-is for now and optimize later?

packages/mcp-server-supabase/src/pricing.ts
gregnr: Yep munch better 🤦‍♂️

packages/mcp-server-supabase/src/server.ts
gregnr: Reverting this. During testing the LLM omitted the `name` property which should have defaulted the value to this default (`develop`), but it didn't - instead we got an API error saying `name` was missing. I removed the default originally to encourage the LLM not to omit it, but I realize now this is actually a zod implementation issue. I'll open a new PR for that.

packages/mcp-server-supabase/src/server.ts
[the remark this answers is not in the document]
gregnr: I'll revert for now and we can address in another PR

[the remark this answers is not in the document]
sweatybridge: Oo ok, I will fix it on the api side instead. The git branch should just be null.

packages/mcp-server-supabase/src/server.ts
sweatybridge: Hmm ok, it might be a bit of vibe killer but we don't have enough data right now to know for sure. I will keep monitoring the number of branches created to see if it becomes a problem.

sweatybridge (approved):

PR #37, opened 2025-04-07
refactor: more explicit tests

gregnr:
Removes global mock orgs and projects and instead moves them to each test. This is slightly more verbose but makes testing much clearer and less error prone since each test is responsible for adding mock orgs & projects in the exact states they require prior to performing the test.

For example, we have a bunch of tests that make sure the cost confirmation logic works correctly depending on what payment plan the org is in. It's a lot clearer to define these mock orgs on a per-test basis than globally.

### DIFF
--- a/packages/mcp-server-supabase/vitest.config.ts
+++ b/packages/mcp-server-supabase/vitest.config.ts
@@ -3,6 +3,6 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   test: {
     include: ['src/**/*.{test,spec}.ts'],
-    testTimeout: 20_000, // PGlite can take a while to initialize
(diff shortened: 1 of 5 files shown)


packages/mcp-server-supabase/vitest.config.ts
```diff
@@ -3,6 +3,6 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   test: {
     include: ['src/**/*.{test,spec}.ts'],
-    testTimeout: 20_000, // PGlite can take a while to initialize
+    testTimeout: 30_000, // PGlite can take a while to initialize
```
gregnr, line 6: Had a few tests that failed due to timeout the fixed upon re-run. Bumping this to 30s.

olirice (approved):

PR #43, opened 2025-04-08
fix: remove pg-meta dep by embedding required logic

gregnr:
A number of Windows-related issues are due to a build issue on `libpg-query`, which is a nested dependency of pg-meta.

This PR cherry picks the logic we need from pg-meta (list tables & list extensions) and embeds it into this server so that we no longer need to depend on pg-meta. In the future, we will replace this logic completely with [supastructure](https://github.com/olirice/supastructure).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -527,108 +527,6 @@
         "node": ">=18"
       }
     },
-    "node_modules/@fastify/ajv-compiler": {
-      "version": "3.6.0",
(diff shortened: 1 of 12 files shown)


olirice (approved):

sweatybridge (approved):

PrathameshRR:
image

it worked. after the merge.

Thank you :')

PR #58, opened 2025-04-17
feat: list and deploy edge functions

gregnr:
Adds Edge Function support via 2 new tools:
- `list_edge_functions`: List all Edge Functions for a project including source files for each.
- `deploy_edge_function`: Deploys an Edge Function. If the specified slug is unique, a new Edge Function will be deployed. Otherwise this will deploy a new version of an existing function.

For now we opted to include source files (eg. `index.ts`) in the `list_edge_functions` call vs. create another detailed `get_edge_function` tool to reduce the number of tools needed. The rationale is that many functions end up living in only a single `index.ts` file, so the overhead of pulling this for every function is not excessive for most people. This might change in the future, but this is a good starting point and can keep tool count down.

### DIFF
--- a/packages/mcp-server-supabase/src/eszip.ts
+++ b/packages/mcp-server-supabase/src/eszip.ts
@@ -1,0 +1,108 @@
+import { build, Parser } from '@deno/eszip';
+import { join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { z } from 'zod';
+
+const parser = await Parser.createInstance();
+const sourceMapSchema = z.object({
+  version: z.number(),
+  sources: z.array(z.string()),
+  sourcesContent: z.array(z.string()).optional(),
+  names: z.array(z.string()),
+  mappings: z.string(),
+});
+
+/**
+ * Extracts source files from an eszip archive.
+ *
+ * Optionally removes the given path prefix from file names.
+ *
+ * If a file contains a source map, it will return the
+ * original TypeScript source instead of the transpiled file.
+ */
+export async function extractFiles(
+  eszip: Uint8Array,
+  pathPrefix: string = '/'
+) {
+  let specifiers: string[] = [];
+
+  if (eszip instanceof ReadableStream) {
+    const reader = eszip.getReader({ mode: 'byob' });
+    specifiers = await parser.parse(reader);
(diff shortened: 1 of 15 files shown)


packages/mcp-server-supabase/src/eszip.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  return files;
+}
+
+/**
+ * Bundles files into an eszip archive.
+ *
+ * Optionally prefixes the file names with a given path.
+ */
+export async function bundleFiles(files: File[], pathPrefix: string = '/') {
```
laktek, line 82: @gregnr what is this used for?

packages/mcp-server-supabase/src/server.ts
```diff
@@ -382,6 +391,146 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
...
+
+          return edgeFunctions;
+        },
+      }),
+      deploy_edge_function: tool({
+        description: `Deploys an Edge Function to a Supabase project. If the function already exists, this will create a new version. Example:\n\n${edgeFunctionExample}`,
+        parameters: z.object({
+          project_id: z.string(),
+          name: z.string().describe('The name of the function'),
+          entrypoint_path: z
```
laktek, line 476: can we also support the `import_map_path`? Can be defaulted to `deno.json`.

packages/mcp-server-supabase/src/eszip.ts
gregnr: Just for tests - I bundle the files into a eszip archive within my mock edge functions so that I can validate that my eszip parsing logic works later when fetching them back (via `/v1/projects/{ref}/functions/{function_slug}/body`)

packages/mcp-server-supabase/src/server.ts
gregnr: Import map path now supported. Logic is:
1. Use `import_map_path` if specified
2. Otherwise fallback to `import_map_path` on existing function if it exists
3. Otherwise use a filename heuristic to guess the path (`deno.json` first, then `import_map.json`)
4. Otherwise leave undefined

olirice (approved):

PR #89, opened 2025-06-06
Storage tools

Ngineer101:
This PR adds 3 new storage tools:

- `list_storage_buckets`: Lists all storage buckets in a Supabase project.
- `get_storage_config`: Gets the storage config for a Supabase project.
- `update_storage_config`: Updates the storage config for a Supabase project (requires a paid plan).

LLMs can use this to get all the buckets in a project and access the storage config (max file upload size, image transformation enabled, etc.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6427,7 +6427,7 @@
     },
     "packages/mcp-server-supabase": {
       "name": "@supabase/mcp-server-supabase",
-      "version": "0.4.2",
+      "version": "0.4.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@deno/eszip": "^0.84.0",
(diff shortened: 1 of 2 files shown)


gregnr:
Hey @Ngineer101 thanks for the PR! Just want to acknowledge that I've seen this.

One tradeoff we've been trying to balance with Supabase MCP is number of tools vs most needed tools. In an ideal world we have as many tools as needed to interact with Supabase in a modular way, but unfortunately we need to deal with tool limits - Cursor for example maxes at 40 tools across all MCP servers. The more tools we introduce the more we creep into this limit.

All that to say I would love to get storage tools in here, but will need to weight this against other tools (like security advisors which we are also working on). Open to ideas - something like this might help: https://github.com/supabase-community/supabase-mcp/issues/52#issuecomment-2802200789

Ngineer101:
@gregnr, sounds good. I'll give it a shot and make a PR soon.

PR #90, opened 2025-06-06
feat(tools): add search_docs tool

charislam:
Add a search_docs tool that integrates with the Docs Content API to provide access to up-to-date Supabase documentation


charislam:
Nice, thanks for the edits @gregnr! Tested on a bunch of queries and seems to call this tool when reasonable (including for auth instructions where _thank goodness_ we finally have something not suggesting outdated auth code 🙌🏼)

PR #103, opened 2025-06-23
Conditional tools based on platform implementation

gregnr:
Supabase MCP supports custom [platform implementations ](https://github.com/supabase-community/supabase-mcp/pull/78) based on the environment it is running in. This PR makes this even more flexible by allowing you to implement only a subset of platform operations.

Before:

```ts
const myPlatform: SupabasePlatform = {
  executeSql() { ... },
  listMigrations() { ... },
  applyMigration() { ... },
  // every other possible platform operations
}
```

After:

```ts
const myPlatform: SupabasePlatform = {
  database: {
    executeSql() { ... },
    listMigrations() { ... },
    applyMigration() { ... },
  },
  // intentionally don't implement anything else
}

### Why
Self-hosted Supabase can support some but not all platform operations (e.g. it can't support account level or branching operations). In order to make this MCP server [compatible with self-hosted](https://github.com/supabase-community/supabase-mcp/pull/92), we need a way to conditionally implement platform operations.

### How
Now that we support [feature groups](https://github.com/supabase-community/supabase-mcp/pull/98), we can extend this logic to also filter by platform operations. The `SupabasePlatform` interface is now grouped by feature (`database`, `account`, `functions`, etc) and each group is optional. We add logic in the server that conditionally adds tools only if the platform methods for that group are implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,9 +156,9 @@ You can enable or disable specific tool groups by passing the `--features` flag 
 npx -y @supabase/mcp-server-supabase@latest --features=database,docs
 ```
 
-Available groups are: [`account`](#account), [`docs`](#knowledge-base), [`database`](#database), [`debug`](#debug), [`development`](#development), [`functions`](#edge-functions), [`storage`](#storage), and [`branching`](#branching-experimental-requires-a-paid-plan).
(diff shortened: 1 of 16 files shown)


olirice (approved):

madhavpujara-at:
any updates on when will this be merged

Issue #112, opened 2025-07-19
Exclude any mutating tools when in read_only mode


gregnr:
Currently when setting the server to read_only mode, this only applies to database operations:

* `execute_sql` (by running as read only postgres user)
* `apply_migration` (by rejecting the tool)

We should change this to also reject the following mutating tools:

* `create_project`
* `pause_project`
* `restore_project`
* `deploy_edge_function`
* `create_branch`
* `delete_branch`
* `merge_branch`
* `reset_branch`
* `rebase_branch`
* `update_storage_config`

We should also update the readme accordingly.

gregnr:
Thanks @sourman. Currently when `read_only` mode is enabled, the tools are still visible to the LLM but are immediately rejected when called. Keeping the tools visible helps the LLM understand what is still possible so that they can prompt the user to adjust `read_only` mode if relevant. If you know for a fact that you don't want them, your approach works great!

Closing this issue now that #114 is complete.

PR #114, opened 2025-07-20
Disable mutating tools in read-only mode

diksipav:

Improvement: #112 

Read-only mode executes SQL as a read-only Postgres user (`execute_sql` tool), and 
`apply_migration` tool is disabled in read-only mode.

## What is the new behavior?

Alongside `execute_sql` and `apply_migration` all other mutating tools are disabled in read-only mode.


packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -7,9 +7,10 @@ import { hashObject } from '../util.js';
 
 export type AccountToolsOptions = {
   platform: SupabasePlatform;
+  readOnly?: Boolean;
```
gregnr, line 10:
```suggestion
  readOnly?: boolean;
```

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -8,11 +8,13 @@ import { injectableTool } from './util.js';
 export type BranchingToolsOptions = {
   platform: SupabasePlatform;
   projectId?: string;
+  readOnly?: boolean;
```
gregnr, line 11: We'll need to update `server.ts` to forward the read only setting to all of these new `readOnly` options

gregnr (changes requested):
Thanks @diksipav 👍 can we please also add some tests for each of these to validate that the errors are thrown in read only mode?

diksipav:
Hey @gregnr it seems to me that for testing a real `ANTHROPIC_API_KEY` is used and we are calling the LLM, right? Also one of the latest (and most expensive models is used). I am testing now with my own api key, but maybe we can at least test with a cheaper model? Also not sure in this case should I test every option or for example to test one tool per each feature group in read-only mode.

gregnr (commented):
@diksipav I think there is no need for e2e tests in this case, let's just stick to unit. The logic is pretty straightforward and probably not worth the cost of running through a real LLM.

The tests you've added to `server.test.ts` look good.

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -377,6 +378,41 @@ describe('tools', () => {
     });
   });
 
+  test('try creating a project in read-only mode', async () => {
```
gregnr, line 381: Test names should include the expected outcome for non-happy paths:

```suggestion
  test('create project in read-only mode throws an error', async () => {
```

packages/mcp-server-supabase/test/mocks.ts
```diff
@@ -179,7 +179,6 @@ export const mockManagementApi = [
       name: z.string(),
       region: z.string(),
       organization_id: z.string(),
-      db_pass: z.string(),
```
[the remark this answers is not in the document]
diksipav: Oh yes, great. I already put it back in the last commit : )

gregnr (approved):

gregnr:
Looks good @diksipav 👍 you can ignore the failed E2E tests since these are just due to the missing API key. We'll merge this soon.

MariusWilsch:
I was wondering why it is not possible to still apply a migration in read-only mode with confirmation, while limiting read-only in execute SQL. 

It's kind of annoying because I want to use a migration system with MCP, but it's not possible in the read-only mode.

gregnr:
Thanks again @diksipav and sorry for the long delay. Merging!

PR #130, opened 2025-08-27
feat: make region required for create_project

mattrossman:
Supporting AI-109


packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -381,7 +381,7 @@ describe('tools', () => {
     });
   });
 
-  test('create project chooses closest region when undefined', async () => {
+  test('create project without region fails', async () => {
```
mattrossman, line 384: I modified the old test but could potentially remove the test since the schema ensures this param is provided and we don't test like this in other cases

PR #141, opened 2025-09-09
feat: mcp registry

gregnr:
Adds the relevant config files and scripts to publish our MCP server to the [official MCP registry](https://blog.modelcontextprotocol.io/posts/2025-09-08-mcp-registry-preview/).
- `server.json`
- script to login to the MCP registry (requires https://github.com/supabase/supabase/pull/38561)
- script to update `server.json` versions to be the same as the `package.json` version
- NPM script aliases for the above


packages/mcp-server-supabase/server.json
```diff
@@ -0,0 +1,62 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
+  "name": "com.supabase/mcp",
```
gregnr, line 3: Would like some ideas for this name. Put as `com.supabase/mcp` as a placeholder. It's nice and simple, but if we ever offer more MCP servers in the future, it'd be nice to think now about a naming convention. Also unsure if "mcp" is redundant here since this entire JSON file is designed for MCP. Another idea is `com.supabase/platform`.

Looking at some [other servers](https://registry.modelcontextprotocol.io/docs#/operations/list-servers) already published, I'm seeing a large variation in naming. Did see a number of `com.example/mcp` though.

gregnr: Sticking with `com.supabase/mcp` 👍

PR #155, opened 2025-10-07
feat: `get_logs` platform-specific sql and 24 hour logs

mattrossman:
- Updates platform signature for `getLogs` to use `service` instead of `sql` option to support the various querying requirements across platforms
- Updates default timestamp to read logs from the past 24 hours instead of 1 minute, and removes the 5 minute special case for branch actions

Resolves [AI-115](https://linear.app/supabase/issue/AI-115)
Resolves [AI-165](https://linear.app/supabase/issue/AI-165)


gregnr (approved):
LGTM

PR #165, opened 2025-10-22
fix(list_table): preventing sql injection attack on list_tables

Rodriguespn:

Security fix - SQL injection vulnerability prevention

The `list_tables` tool was vulnerable* to SQL injection via the `schemas` parameter due to direct string interpolation in SQL queries.

*Note the vulnerability isn't quite as bad as it sounds:
1. MCP clients already have full SQL access via `execute_sql` tool call (by design), so attempting to inject SQL in `list_tables` doesn't escalate privileges (you already have full SQL permissions). 
2. `list_tables` is restricted to a readonly postgres role, so injected SQL can't mutate the database

But this is still unexpected behavior, especially if you intentionally disabled the `execute_sql` tool on the client side assuming that this adds security.

**Related Issue:** https://linear.app/supabase/issue/AI-139/sql-injection-in-list-tables-supabase-mcp-server-via-schemas-parameter

## What is the new behavior?

We [added](https://github.com/supabase/postgres-meta/pull/1001) parameterized query support to the downstream `/query` route on postgres-meta. This PR updates the MCP server to use this new feature.

- **Query Structure**: Changed `listTablesSql()` to return `{query, parameters}` with PostgreSQL placeholders (`$1`, `$2`, etc.)
- **Type Safety**: Added `parameters` field to `executeSqlOptionsSchema` 
- **Platform Layer**: Updated `executeSql()` to pass parameters through API
- **Tool Layer**: Modified `list_tables` to use parameterized queries
- **Test Infrastructure**: Enhanced HTTP mocks to support both parameterized and plain queries
- **Security Test**: Added test confirming malicious input is treated as literal value

## Additional context

All 123 unit tests passing. SQL injection test verifies that malicious schema inputs like `"public'; DROP TABLE public.sqli_test; --"` are safely handled as literal values, returning empty results without executing injected SQL.

### DIFF
--- a/packages/mcp-server-supabase/src/pg-meta/index.ts
+++ b/packages/mcp-server-supabase/src/pg-meta/index.ts
@@ -36,9 +36,6 @@ export function listTablesSql(schemas: string[] = []) {
     sql += `where schema not in (${placeholders})`;
     parameters = SYSTEM_SCHEMAS;
   }
-  console.log('Schema', schemas);
-  console.log('Generated SQL:', sql);
-  console.log('With parameters:', parameters);
 
   return { query: sql, parameters };
 }
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -1092,8 +1092,6 @@ describe('tools', () => {
       },
     });
 
-    console.log('Malicious Result:', maliciousResult);
-
     // Should return empty array without errors, proving the SQL injection was prevented
     expect(maliciousResult).toEqual([]);
   });


packages/mcp-server-supabase/src/platform/types.ts
```diff
@@ -110,6 +110,7 @@ export const deployEdgeFunctionOptionsSchema = z.object({
 
 export const executeSqlOptionsSchema = z.object({
   query: z.string(),
+  parameters: z.array(z.any()).optional(),
```
gregnr: Safer to use `unknown` here (this is also how we type it in postgres-meta/infra)

```suggestion
  parameters: z.array(z.unknown()).optional(),
```

packages/mcp-server-supabase/test/mocks.ts
```diff
@@ -258,7 +258,10 @@ export const mockManagementApi = [
   /**
    * Execute a SQL query on a project's database
    */
-  http.post<{ projectId: string }, { query: string; read_only?: boolean }>(
+  http.post<
+    { projectId: string },
+    { query: string; parameters?: any[]; read_only?: boolean }
```
gregnr:
```suggestion
    { query: string; parameters?: unknown[]; read_only?: boolean }
```

packages/mcp-server-supabase/test/mocks.ts
```diff
@@ -269,30 +272,43 @@ export const mockManagementApi = [
...
-      const statementResults = await db.exec(wrappedQuery);
-
-      // Remove last result, which is for the "RESET ROLE" statement
-      statementResults.pop();
-
-      const lastStatementResults = statementResults.at(-1);
+      // Set role before executing query
+      await db.exec(
+        `SET ROLE ${read_only ? 'supabase_read_only_role' : 'postgres'};`
+      );
```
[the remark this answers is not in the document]
Rodriguespn: Nice call @gregnr 🙌 I've swapped the db.exec() for tx.exec() after creating a transaction

packages/mcp-server-supabase/src/pg-meta/index.ts
```diff
@@ -25,14 +25,22 @@ export function listTablesSql(schemas: string[] = []) {
...
+    parameters = schemas;
   } else {
-    sql += `where schema not in (${SYSTEM_SCHEMAS.map((s) => `'${s}'`).join(',')})`;
+    const placeholders = SYSTEM_SCHEMAS.map((_, i) => `$${i + 1}`).join(', ');
+    sql += `where schema not in (${placeholders})`;
+    parameters = SYSTEM_SCHEMAS;
   }
+  console.log('Schema', schemas);
+  console.log('Generated SQL:', sql);
+  console.log('With parameters:', parameters);
```
mattrossman:
```suggestion
```

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -1060,6 +1060,44 @@ describe('tools', () => {
...
+    const maliciousResult = await callTool({
+      name: 'list_tables',
+      arguments: {
+        project_id: project.id,
+        schemas: [maliciousSchema],
+      },
+    });
+
+    console.log('Malicious Result:', maliciousResult);
+
```
mattrossman:
```suggestion
```

PR #166, opened 2025-10-23
feat: lazy graphql schema

mattrossman:
Updates server to only fetch the GraphQL schema for `search_docs` description once per session, even for stateless transports.

**Changes**
- `validateSchema` now defaults to `false`, disabling local schema validation in favor of server-side validation with comparable error reporting verified in https://github.com/supabase-community/supabase-mcp/pull/166#issuecomment-3444349514
- Tool `description` changed from `string` to `Prop<string>` allowing sync/async callbacks to lazily construct descriptions as-needed during `"tools/list"` requests
- New test `"schema is only loaded when listing tools"` checks that the (mocked) schema loading only occurs when listing tools and not on tool calls

Resolves [AI-179](https://linear.app/supabase/issue/AI-179/optimize-content-api-client-creation-to-reduce-traffic-spikes)

### DIFF
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -11,6 +11,7 @@ import {
   ACCESS_TOKEN,
   API_URL,
   contentApiMockSchema,
+  mockContentApiSchemaLoadCount,
   createOrganization,
   createProject,
   createBranch,
@@ -31,6 +32,7 @@ beforeEach(async () => {
   mockOrgs.clear();
   mockProjects.clear();
   mockBranches.clear();
+  mockContentApiSchemaLoadCount.value = 0;
 
   const server = setupServer(...mockContentApi, ...mockManagementApi);
   server.listen({ onUnhandledRequest: 'error' });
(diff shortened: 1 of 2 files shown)


packages/mcp-server-supabase/src/tools/docs-tools.ts
```diff
@@ -10,14 +10,16 @@ export type DocsToolsOptions = {
...
+      description: async () => {
+        const schema = await contentApiClient.loadSchema();
 
-        You should default to calling this even if you think you already know the answer, since the documentation is always being updated.
-
-        Below is the GraphQL schema for the Supabase docs endpoint:
-        ${contentApiClient.schema}
-      `,
+        return source`
+          Search the Supabase documentation using GraphQL. Must be a valid GraphQL query.
```
Rodriguespn: Not related with this PR but I've noticed that this tool is only called when I start my prompts with "How can I... or How do I..." (at least for me). Sometimes, if it bumps into any problem while calling other tool, the LLM will search for the docs online instead of calling this tool.

If anyone else is experiencing this, my suggestion is:
1. We write explicitly in the tool description "If you face any issue while calling other Supabase tool, use the search_docs tool to get the most up-to-date documentation to help you debug"
2. We put this disclaimer on every tool description

I prefer the first one because we only have to maintain this instruction once.

PS: I would only change this behavior once we have evals in place

PR #169, opened 2025-10-27
feat: Add MCP elicitation for create project and branch tools

Rodriguespn:
## Summary

Implements MCP elicitation for the `apply_migration` tool to request user confirmation before applying database migrations.

## Implementation

Uses a try-catch pattern to gracefully fallback when clients don't support elicitation:
- **Supporting clients**: Show confirmation dialog with migration details and SQL preview
- **Non-supporting clients**: Proceed with migration without confirmation (existing behavior)

## Client Support

Currently supported: VS Code Copilot, Cursor, Postman, VT Code, fast-agent, mcp-use, MCPJam

Not yet supported: Claude Desktop, Claude.ai, GitHub Copilot coding agent

## Related

[AI-142: Explore MCP Elicitations](https://linear.app/supabase/issue/AI-142)

### DIFF
--- a/packages/mcp-server-supabase/src/tools/account-tools.ts
+++ b/packages/mcp-server-supabase/src/tools/account-tools.ts
@@ -1,5 +1,6 @@
 import { tool, type ToolExecuteContext } from '@supabase/mcp-utils';
 import type { Server } from '@modelcontextprotocol/sdk/server/index.js';
+import { ElicitResultSchema } from '@modelcontextprotocol/sdk/types.js';
 import { z } from 'zod';
 import type { AccountOperations } from '../platform/types.js';
 import { type Cost, getBranchCost, getNextProjectCost } from '../pricing.js';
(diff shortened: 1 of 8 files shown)


packages/mcp-server-supabase/src/tools/database-operation-tools.ts
```diff
@@ -213,6 +216,47 @@ export function getDatabaseTools({
           throw new Error('Cannot apply migration in read-only mode.');
         }
 
+        // Try to request user confirmation via elicitation
+        if (server) {
+          try {
+            const result = (await server.request(
```
mattrossman: Wondering if we can use the [elicitInput](https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#eliciting-user-input) helper here

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -37,25 +37,73 @@ export function getBranchingTools({
...
+        // branch cost interactively and this parameter is not required. For clients
+        // without elicitation support, this confirmation ID is required.
         confirm_cost_id: z
-          .string({
-            required_error:
-              'User must confirm understanding of costs before creating a branch.',
-          })
-          .describe('The cost confirmation ID. Call `confirm_cost` first.'),
+          .string()
+          .optional()
```
mattrossman: Wonder if we can make the schema / tool implementation dynamic? For clients that don't support elicitation, the field isn't really optional.

mattrossman:
CleanShot 2025-10-31 at 09 58 35@2x

I think we can filter out `get_cost` too when elicitations are supported, that's mainly a supporting tool for `confirm_cost`

gregnr (commented):
Since elicitations are a major new feature, maybe it'd be best to implement them in just one place first and iterate in future PRs? e.g. this PR could only focus on `apply_migration` since that is relatively straightforward.

packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -76,11 +83,16 @@ export function getAccountTools({ account, readOnly }: AccountToolsOptions) {
         return await account.getProject(id);
       },
     }),
-    get_cost: tool({
-      description:
-        'Gets the cost of creating a new project or branch. Never assume organization as costs can be different for each.',
+    get_and_confirm_cost: tool({
```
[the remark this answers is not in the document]
mattrossman: In practice having the elicitation in `create_project` felt odd to me because you have to approve a `create_project` tool call before knowing how much the project will cost or that a safety net will follow ([Slack context](https://supabase.slack.com/archives/C051L8U2EJF/p1761920640958849?thread_ts=1761846825.806379&cid=C051L8U2EJF)).

PR #187, opened 2025-11-25
feat: minifiy content api graphql schema

gregnr:
[Minifies](https://github.com/drwpow/gqlmin) our content API's GraphQL schema before injecting it into the `search_docs` tool description. This strips out insignificant whitespace and removes comments, reducing the character size by 75% (actual token savings are model-dependant but should be similar, e.g. ~76% for [GPT tokenizer](https://platform.openai.com/tokenizer)).

This helps lower the amount of context that MCP clients need to load for their LLMs (translating to cost savings).

### Are there downsides?
The main downside is stripping out comments, which usually benefit LLMs by giving them a richer description of how to use the API. I tested the minified version with Cursor and it successfully formed the GraphQL query with no errors:

image

I think the schema is pretty self-describing without the comments, so the benefit of removing the context is likely worth it.

### DIFF
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -5,6 +5,7 @@ import {
 } from '@modelcontextprotocol/sdk/types.js';
 import { StreamTransport } from '@supabase/mcp-utils';
 import { codeBlock, stripIndent } from 'common-tags';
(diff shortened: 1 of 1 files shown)


olirice (approved):

mattrossman (approved):
Tested in Cursor, drops Composer context usage a couple %

PR #194, opened 2025-12-17
fix(content api): use get requests

charislam:
Use GET requests instead of POST for Content API to cache responses if possible.

Also fixes a bug in the tests where mock servers weren't being correctly cleaned up. Switching from POST to GET exposed this bug because it allowed a single request to be handled multiple times by multiple mock servers.

Tested locally with Cursor.

Resolves AI-318


mattrossman (approved):
Tested working on my end as well

PR #198, opened 2025-12-29
feat: Zod v4, AI SDK v6

mattrossman:
**Changes:**
- Adds common deps (Zod, MCP SDK, AI SDK and supporting packages) to pnpm `catalog:` to ensure version sync across `packages/`
- Adds `--pnpm` flag to pkg-pr-new to ensure `catalog:` references correctly resolve before preview publish
- Marks Zod as a peer dependency in range `^3.25.0 || ^4.0.0` per their [library authors guidance](https://zod.dev/library-authors#how-to-configure-peer-dependencies) and includes the latest version in our dev dependencies, so consumers can "bring their own Zod"
- Updates Zod imports to `"zod/v4"` to ensure usage of v4 APIs
- Updates AI SDK and related packages to latest versions (v6) and fixes breaking changes (e.g. maxSteps -> stepCountIs())
- Updates MCP SDK to the oldest version that supports Zod 4 (1.23.0) to prevent introducing changes from newer protocol in subsequent versions
- Migrates deprecated Zod `required_error` and `errorMap` fields to unified `error` field
- Removes `zod-to-json-schema` dependency in favor of Zod v4 native `.toJSONSchema()`
- **Breaking:** replaces `currentFeatureGroupSchema` export with `CURRENT_FEATURE_GROUPS` to avoid v3/v4 Zod schema conflicts (consumers should instead create the enum schema themselves from the string array)

I tested the [preview release](https://github.com/supabase-community/supabase-mcp/pull/198#issuecomment-3697368278) works (after small tweak to use `CURRENT_FEATURE_GROUPS` instead of the previously exported schema) in platform repo. The supabase monorepo doesn't consume schemas from this package.

Resolves [AI-313](https://linear.app/supabase/issue/AI-313/update-mcp-server-package-to-support-zod-4), [AI-333](https://linear.app/supabase/issue/AI-333/upgrade-mcp-packages-to-vercel-ai-sdk-6)

### DIFF
--- a/packages/mcp-server-supabase/package.json
+++ b/packages/mcp-server-supabase/package.json
@@ -59,7 +59,7 @@
     "openapi-fetch": "^0.13.5"
   },
   "peerDependencies": {
-    "zod": "^3.25.0 || ^4.0.0"
+    "zod": "catalog:"
   },
   "devDependencies": {
     "@ai-sdk/anthropic": "catalog:",
--- a/packages/mcp-utils/package.json
+++ b/packages/mcp-utils/package.json
@@ -32,7 +32,7 @@
     "@modelcontextprotocol/sdk": "catalog:"
   },
   "peerDependencies": {
-    "zod": "^3.25.0 || ^4.0.0"
+    "zod": "catalog:"
   },
   "devDependencies": {
     "@total-typescript/tsconfig": "^1.0.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -19,7 +19,7 @@ catalogs:
       specifier: ^6.0.3
       version: 6.0.3
     zod:
-      specifier: ^4.2.1
+      specifier: ^3.25.0 || ^4.0.0
       version: 4.2.1
 
 importers:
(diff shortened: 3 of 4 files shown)


mattrossman:
Note on Zod v4 imports. From [How to support Zod 4:](https://zod.dev/library-authors?id=how-to-support-zod-4)

> "zod/v4" and "zod/v4/mini"— ❌ These subpaths are the homes of Zod 4 Classic and Mini, respectively. If you want your library to work with both Zod and Zod Mini, you should build against the base classes defined in "zod/v4/core". If you reference classes from the "zod/v4" module, your library will not work with Zod Mini, and vice versa. This is extremely discouraged. Use "zod/v4/core" instead, which exports the $-prefixed subclasses that are extended by Zod Classic and Zod Mini. The internals of the classic & mini subclasses are identical; they only differ in which helper methods they implement.

Although this guide recommends importing from `zod/v4/core` instead of `zod/v4`, supporting Zod Mini isn't relevant for our use case, and [Zod Core](https://zod.dev/packages/core) is much lower level in the APIs offered which seems like unnecessary effort.

PR #222, opened 2026-02-20
fix: replace deprecated Anthropic model IDs in tests

mattrossman:
Replaces retired Anthropic model IDs in e2e tests:
- `claude-3-7-sonnet-20250219` → `claude-sonnet-4-6`
- `claude-3-5-haiku-20241022` → `claude-haiku-4-5-20251001`

Also adjusts a `toMatchCriteria` assertion in the edge function deploy test that was consistently failing with Sonnet 4.6's stricter judge ("describes steps to test" → "includes an example of how to invoke").

Closes AI-444

### DIFF
--- a/packages/mcp-server-supabase/test/e2e/functions.e2e.ts
+++ b/packages/mcp-server-supabase/test/e2e/functions.e2e.ts
@@ -57,7 +57,7 @@ describe('edge function e2e tests', () => {
     );
 
     await expect(text).toMatchCriteria(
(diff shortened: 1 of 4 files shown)


gregnr (approved):

Rodriguespn (approved):
Approved ✅ 
Not a blocker but having the same model name hard coded in different files made me think if we shouldn't have a constant file where we store all hard coded strings

PR #244, opened 2026-03-24
chore: restrict local paths edge_functions

staaldraad:

chore

mcp server has  lax  local path name acceptance

## What is the new behavior?

enforces paths to be in the current working directory

### DIFF
--- a/packages/mcp-server-supabase/src/tools/edge-function-tools.ts
+++ b/packages/mcp-server-supabase/src/tools/edge-function-tools.ts
@@ -6,6 +6,7 @@ import {
   edgeFunctionWithBodySchema,
 } from '../platform/types.js';
 import { injectableTool, type ToolDefs } from './util.js';
+import path from 'path';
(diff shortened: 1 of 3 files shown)


packages/mcp-server-supabase/src/tools/edge-function-tools.ts
```diff
@@ -48,7 +49,17 @@ const deployEdgeFunctionInputSchema = z.object({
   files: z
     .array(
       z.object({
-        name: z.string(),
+        name: z.string().refine(
+          (filePath) => {
+            const resolved = path.resolve(process.cwd(), filePath);
```
mattrossman, line 54: This looks off to me—in this context, name represents the path within the edge function bundle where the function will be written, not a file path on the machine running the MCP process. Why would `process.cwd()` be a factor? I might need to see a test case showing what sort of scenario this prevents to better understand the motivation.

PR #245, opened 2026-03-24
chore: path normalization

staaldraad:

Bug fix

Path is not normalized before concatting to apiUrl, leading to malformed paths.

## What is the new behavior?

Path is normalized

### DIFF
--- a/packages/mcp-server-postgrest/src/server.ts
+++ b/packages/mcp-server-postgrest/src/server.ts
@@ -75,7 +75,9 @@ export function createPostgrestMcpServer(options: PostgrestMcpServerOptions) {
             .optional(),
         }),
         async execute({ method, path, body }) {
-          const url = new URL(`${apiUrl}${path}`);
+          // normalize path concating to apiUrl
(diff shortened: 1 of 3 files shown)


packages/mcp-server-postgrest/src/server.ts
```diff
@@ -75,7 +75,9 @@ export function createPostgrestMcpServer(options: PostgrestMcpServerOptions) {
             .optional(),
         }),
         async execute({ method, path, body }) {
-          const url = new URL(`${apiUrl}${path}`);
+          // normalize path concating to apiUrl
+          const normalizedPath = new URL(path, 'http://mock/').pathname;
+          const url = new URL(`${apiUrl}${normalizedPath}`);
```
mattrossman, line 81: Looks like we're dropping query params here, is that expected?

Would be helpful to have some test case(es) showing what scenarios this prevents against.

mattrossman (approved):

mattrossman:
Verified new test passes, thanks!

PR #251, opened 2026-04-07
feat: inject RLS advisory into list_tables response

myerekapan:
## Summary

- When `list_tables` returns any table with Row Level Security disabled, appends a `critical` advisory to the response with per-table remediation SQL
- Introduces the shared advisory schema (`id`, `priority`, `level`, `title`, `message`, `remediation_sql`, `doc_url`) and priority-based suppression (`selectAdvisory`) for future GROWTH advisory types
- Filters out system schemas using the same exclusion list as the database linter (`rls_disabled_in_public`)

## Response shape

When RLS is disabled on any user-schema table:

```json
{
  "tables": [
    { "name": "public.users", "rls_enabled": false, "rows": 42 },
    { "name": "public.posts", "rls_enabled": true, "rows": 100 }
  ],
  "advisory": {
    "id": "rls_disabled",
    "priority": 1,
    "level": "critical",
    "title": "Row Level Security is disabled",
    "message": "1 table(s) do not have Row Level Security (RLS) enabled: public.users...",
    "remediation_sql": "ALTER TABLE public.users ENABLE ROW LEVEL SECURITY;",
    "doc_url": "https://supabase.com/docs/guides/database/postgres/row-level-security"
  }
}
```

When all tables have RLS enabled, the `advisory` field is omitted entirely.

## Advisory coordination

This is priority 1 (security, highest) and establishes the schema for future GROWTH tasks.

Max 1 advisory per response. `selectAdvisory()` picks the lowest priority number when multiple candidates apply.

Ref: [GROWTH-712](https://linear.app/supabase/issue/GROWTH-712)

## Test plan

- [x] `buildRlsDisabledAdvisory` — advisory generated for RLS-disabled tables
- [x] `buildRlsDisabledAdvisory` — no advisory when all RLS enabled
- [x] `buildRlsDisabledAdvisory` — system schemas (auth, storage, pg_catalog, etc.) ignored
- [x] `buildRlsDisabledAdvisory` — mixed system/user schemas only reports user tables
- [x] `buildRlsDisabledAdvisory` — handles custom user schemas
- [x] `selectAdvisory` — picks highest priority (lowest number)
- [x] `selectAdvisory` — returns null for empty/all-null candidates
- [x] Integration: `list_tables` compact response includes advisory when RLS disabled
- [x] Integration: `list_tables` omits advisory when all tables have RLS enabled
- [x] All 93 existing server tests pass

### DIFF
--- a/packages/mcp-server-supabase/test/e2e/projects.e2e.ts
+++ b/packages/mcp-server-supabase/test/e2e/projects.e2e.ts
@@ -54,7 +54,7 @@ describe('project management e2e tests', () => {
     const toolCalls = steps.flatMap((step) => step.staticToolCalls);
     const text = steps.at(-1)?.text ?? '';
 
-    expect(toolCalls).toHaveLength(2);
+    expect(toolCalls.length).toBeGreaterThanOrEqual(2);
     expect(toolCalls.at(0)?.toolName).toBe('list_projects');
     expect(toolCalls.at(1)?.toolName).toBe('list_tables');
 
--- a/packages/mcp-server-supabase/src/advisories/advisories.test.ts
+++ b/packages/mcp-server-supabase/src/advisories/advisories.test.ts
@@ -1,0 +1,124 @@
+import { describe, expect, test } from 'vitest';
+import { buildRlsDisabledAdvisory } from './rls-disabled.js';
+import { type Advisory, selectAdvisory } from './schema.js';
+
+describe('buildRlsDisabledAdvisory', () => {
+  test('returns advisory when tables have RLS disabled', () => {
+    const tables = [
+      { name: 'public.users', rls_enabled: false },
+      { name: 'public.posts', rls_enabled: true },
+      { name: 'public.comments', rls_enabled: false },
+    ];
+
+    const advisory = buildRlsDisabledAdvisory(tables);
+
+    expect(advisory).not.toBeNull();
+    expect(advisory!.id).toBe('rls_disabled');
+    expect(advisory!.priority).toBe(1);
+    expect(advisory!.level).toBe('critical');
+    expect(advisory!.message).toContain('public.users');
+    expect(advisory!.message).toContain('public.comments');
+    expect(advisory!.message).not.toContain('public.posts');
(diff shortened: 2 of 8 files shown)


packages/mcp-server-supabase/test/e2e/projects.e2e.ts
```diff
@@ -54,7 +54,7 @@ describe('project management e2e tests', () => {
     const toolCalls = steps.flatMap((step) => step.staticToolCalls);
     const text = steps.at(-1)?.text ?? '';
 
-    expect(toolCalls).toHaveLength(2);
+    expect(toolCalls.length).toBeGreaterThanOrEqual(2);
```
myerekapan, line 57: The E2E test sends "What tables do I have?" to AI model and asserts it makes exactly 2 tool calls: list_projects → list_tables.

Before this PR, list_tables returned just { tables: [...] }. Now it also returns an advisory with "RLS is disabled" + remediation SQL. AI agent sees this and sometimes decides to act on it (e.g., calling apply_migration to enable RLS), resulting in a 3rd tool call.

The test isn't testing whether AI agent makes exactly N calls, it's testing that AI agent correctly identifies the project before listing tables. 

So the fix changes toHaveLength(2) to toBeGreaterThanOrEqual(2) while still asserting the first two calls are list_projects and list_tables in order. The LLM reacting to the advisory is actually the intended behavior of this feature. Does this change make sense or do you have another preference here?

mattrossman: That makes sense, I'm also fine with removing that `.length` assertion since as you mentioned we already check that the first two calls are list_projects and list_tables in order below this.

mattrossman (approved):
Sanity checked the preview build works.

Note in my testing the agent (Cursor with "auto" model) didn't make much of a fuss about the critical advisory, it was interpreted as more of a light warning and didn't surface the remediation suggestion. We may want to add evals to verify that not only is the advisory is in the MCP response, but that agents are actually responding to it appropriately.

Image

PR #254, opened 2026-04-15
fix(security): missing maximum operation limits: unbounded file array and content size in deployedgefunction [MEDIUM]

failsafesecurity:
### Security Finding: Missing Maximum Operation Limits: Unbounded file array and content size in deployEdgeFunction

**Severity:** MEDIUM
**Reported by:** FailSafe Security Researcher
**Component:** `packages/mcp-server-supabase/src/platform/types.ts:75`

### Description

The vulnerability identified in the source file involves missing maximum operation limits on file arrays and content sizes. Specifically, the code at the specified source implements the `deployEdgeFunctionOptionsSchema` but fails to adequately enforce `.max()` constraints on the `files` array and the `content` string. This omission allows an attacker or a malfunctioning LLM to provide a massive number of files or extremely large file contents. In the context of deploying an edge function, this means that the MCP server reads this unbounded array and constructs a `FormData` object with `Blob` instances for each file in memory. The root cause is the absence of strict schema validation boundaries. When user-controlled input is processed by this component without proper validation or boundary checks, the system enters an insecure state leading to resource exhaustion. This type of flaw is particularly dangerous because it trivially enables OOM crashes. Furthermore, the lack of defense-in-depth mechanisms exacerbates the risk, potentially leading to severe operational disruption.

### Fix

Update the `deployEdgeFunctionOptionsSchema` in `packages/mcp-server-supabase/src/platform/types.ts` to enforce strict `.max()` limits on both the `files` array and the `content` string using Zod. ```typescript export const deployEdgeFunctionOptionsSchema = z.object({ // ... other fields files: z.array( z.object({ name: z.string().max(255), content: z.string().max(5 * 1024 * 1024) // 5MB limit per file }) ).max(50) // Max 50 files }); ``` Ensure that the limits align with the Supabase Edge Functions deployment constraints.

### DIFF
--- a/packages/mcp-server-supabase/src/platform/types.ts
+++ b/packages/mcp-server-supabase/src/platform/types.ts
@@ -109,10 +109,10 @@ export const deployEdgeFunctionOptionsSchema = z.object({
   verify_jwt: z.boolean().optional(),
   files: z.array(
     z.object({
-      name: z.string(),
-      content: z.string(),
+      name: z.string().max(255),
+      content: z.string().max(5 * 1024 * 1024),
     })
-  ),
+  ).max(50),
 });
 
 export const executeSqlOptionsSchema = z.object({


failsafesecurity:
Closing due to no response after 35+ days. Will consider resubmission via bounty platform. Feel free to reopen if this is still being reviewed.

PR #268, opened 2026-05-01
fix: MCP registry publication in release workflow

mattrossman:
- Specifies per-platform `asset_pattern` for MCP Publisher in `mise.toml` so mise picks the correct release asset (`mcp-publisher_*` instead of `registry_*`)
- Add MCP registry login step before publish using domain HTTP verification (our listing is under `com.supabase`, which requires the `DOMAIN_VERIFICATION_KEY` secret — see `packages/mcp-server-supabase/scripts/registry/login.sh`)
- Add `workflow_dispatch` trigger so the workflow can be manually re-run without a dummy commit (e.g. to recreate a closed release PR)
- Update contributing guide with a Releases section explaining the automated flow, and reframe manual MCP registry publish as an optional fallback, see https://github.com/supabase-community/supabase-mcp/blob/mattrossman/ai-695-fix-mcp-registry-publication-in-release-workflow/CONTRIBUTING.md#releases

I've populated `DOMAIN_VERIFICATION_KEY` in repo secrets.

**Verification**

I created (and reverted) a temp test workflow that installs `mcp-publisher` and runs `mcp-publisher --version`, see that successful run here.

CleanShot 2026-05-01 at 14 29 35@2x

https://github.com/supabase-community/supabase-mcp/actions/runs/25227199423/job/73973577890?pr=268

References:
- https://modelcontextprotocol.io/registry/github-actions

Closes AI-695


.github/workflows/release.yml
```diff
@@ -116,6 +120,16 @@ jobs:
...
 
+      - name: Install mcp-publisher
+        if: ${{ steps.release.outputs['packages/mcp-server-supabase--release_created'] }}
+        run: |
+          curl -fsSL "https://github.com/modelcontextprotocol/registry/releases/download/v${{ env.MCP_PUBLISHER_VERSION }}/mcp-publisher_linux_amd64.tar.gz" | tar xz mcp-publisher
+          sudo mv mcp-publisher /usr/local/bin/mcp-publisher
+
+      - name: Authenticate to MCP registry
+        if: ${{ steps.release.outputs['packages/mcp-server-supabase--release_created'] }}
+        run: mcp-publisher login github-oidc
```
gregnr: I'm not convinced this will work. Our MCP listing is published under our [domain namespace](https://modelcontextprotocol.info/tools/registry/cli/#ownership-verification) (com.supabase) which requires either DNS or HTTP verification. We will likely need our `domain-verification-key.pem` added as a secret that CI can use to login via the HTTP method.

mise.toml
```diff
@@ -5,4 +5,3 @@ lockfile = true
 [tools]
 node = "lts"
 pnpm = "10"
-"github:modelcontextprotocol/registry" = "latest"
```
gregnr, line 8: A bit sad to see this go as it was one of the main motivators for mise.

> the github: backend was picking the wrong release asset (registry_* instead of mcp-publisher_*)

Is this configurable? I wonder how this worked for us locally?

mattrossman: It's configurable with [asset_pattern](https://mise.jdx.dev/dev-tools/backends/github.html#asset-pattern), it's just a bit verbose if we want to handle [all platforms they release](https://github.com/modelcontextprotocol/registry/releases). 

```toml
[tools."github:modelcontextprotocol/registry"]
version = "latest"

[tools."github:modelcontextprotocol/registry".platforms]
linux-x64 = { asset_pattern = "mcp-publisher_linux_amd64.tar.gz" }
linux-arm64 = { asset_pattern = "mcp-publisher_linux_arm64.tar.gz" }
macos-x64 = { asset_pattern = "mcp-publisher_darwin_amd64.tar.gz" }
macos-arm64 = { asset_pattern = "mcp-publisher_darwin_arm64.tar.gz" }
windows-x64 = { asset_pattern = "mcp-publisher_windows_amd64.tar.gz" }
windows-arm64 = { asset_pattern = "mcp-publisher_windows_arm64.tar.gz" }
```

I figured since it's unlikely devs will need to install/use `mcp-publisher` locally anyway with the automated release setup, I might as well use the install method from MCP docs but I agree the unified setup w/ mise across GitHub actions and local dev was nicer.

I'll switch back to mise since it's not horribly verbose.

mattrossman: ✅ Confirmed working in test run 
https://github.com/supabase-community/supabase-mcp/actions/runs/25229961355/job/73982753688

.github/workflows/release.yml
mattrossman: Good catch, I don't fully understand OIDC but recalled it being set up for NPM publication.

I've added a `DOMAIN_VERIFICATION_KEY` secret to the repo containing the contents of our `domain-verification-key.pem` and updated the workflow to use the `pnpm registry:login` script for domain verification.

mise.toml
```diff
@@ -5,4 +5,14 @@ lockfile = true
...
+[tools."github:modelcontextprotocol/registry"]
+version = "latest"
+
+[tools."github:modelcontextprotocol/registry".platforms]
+linux-x64 = { asset_pattern = "mcp-publisher_linux_amd64.tar.gz" }
+linux-arm64 = { asset_pattern = "mcp-publisher_linux_arm64.tar.gz" }
+macos-x64 = { asset_pattern = "mcp-publisher_darwin_amd64.tar.gz" }
+macos-arm64 = { asset_pattern = "mcp-publisher_darwin_arm64.tar.gz" }
+windows-x64 = { asset_pattern = "mcp-publisher_windows_amd64.tar.gz" }
+windows-arm64 = { asset_pattern = "mcp-publisher_windows_arm64.tar.gz" }
```
gregnr, line 18: I don't suppose `asset_pattern` supports regex/glob patterns? Not sure if there's a catch-all/global platform option anyways though.

gregnr (approved):

mise.toml
mattrossman: It does to some extent,`asset_pattern` supports globs and templates but neither works here cleanly, `glob mcp-publisher_*` still matches multiple assets and mise picks the wrong one for non-current platforms in the lockfile. Template variables like `{{ darwin_os }}` aren't available until install time so mise errors at config load.

gregnr: Makes sense. Not worth it!

PR #269, opened 2026-05-01
fix: tools not loading on stdio server

### LINKED ISSUE #261: MCP tools don't appear in Claude Code
I have an issue that I can't access Supabase MCP through Claude Code. It just keeps saying it doesn't have access to that MCP despite it working from Claude Chat (I can access Supabse there).

I tried both Claude Code Web and Claude Code Desktop.

Same stupid issue everytime.

Reinstalled, removed access tokens, deleted everything and started fresh. Same issue each time.

gregnr:
## Problem
The docs Content API GraphQL response schema (used for the `search_docs` tool) used a discriminated union with `z.undefined()` to model success/error branches. There were 2 issues:

1. Zod v4 had a breaking change where `z.undefined()` on object properties requires the key to be explicitly present (and set as `undefined`). Absent keys failed validation, which broke tool loading
2. The union was actually incorrect from the beginning. The GraphQL spec allows both `data` and `errors` at the same time (partial results)

## Fix

This PR replaces the union with a flat schema. `data` uses `.nullish()` because the spec allows it to be either absent (validation errors before execution) or `null` (resolver errors on non-nullable fields). `errors` uses `.optional()` since it's simply absent on success.

## How to test

Configure Claude Code with the following MCP server config:

```json
{
  "mcpServers": {
    "supabase-stdio": {
      "command": "npx",
      "args": [
        "-y",
        "@supabase/mcp-server-supabase@https://pkg.pr.new/supabase-community/supabase-mcp/@supabase/mcp-server-supabase@bbbe8ba"
      ],
      "env": {
        "SUPABASE_ACCESS_TOKEN": "<pat>"
      }
    }
  }
}
```

Run the `claude` CLI, enter `/mcp`, select `supabase-stdio`.
1. Confirm that the server connected successfully
2. Confirm that tools are available (by the presence of the **View tools** option, which displays all the tools when selected)

---

Resolves #261

### DIFF
--- a/packages/mcp-server-supabase/src/content-api/graphql.ts
+++ b/packages/mcp-server-supabase/src/content-api/graphql.ts
@@ -13,11 +13,6 @@ export const graphqlRequestSchema = z.object({
   variables: z.record(z.string(), z.unknown()).optional(),
 });
 
-export const graphqlResponseSuccessSchema = z.object({
-  data: z.record(z.string(), z.unknown()),
-  errors: z.undefined(),
-});
-
 export const graphqlErrorSchema = z.object({
   message: z.string(),
   locations: z.array(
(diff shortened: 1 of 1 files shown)


mattrossman (approved):
Preview build tested working on my end 👍

Issue #277, opened 2026-05-09
Feature Request: Consolidated Database Schema Documentation Tool


hasithasandunlakshan:
# Feature request: Consolidated Database Schema Documentation Tool

## Is your feature request related to a problem? Please describe.

Currently, when an AI agent (or a new developer) wants to understand the structure of a Supabase project via MCP, it must call multiple tools like `list_tables` and `list_extensions` and then piece together the schema manually. This is inefficient in terms of token usage and context window consumption for LLMs. Furthermore, critical details like Row Level Security (RLS) policies, database triggers, and custom SQL functions are not easily accessible through a single high-level overview, making it difficult to extract a comprehensive snapshot for documentation or onboarding.

## Describe the solution you'd like

I propose adding a new read-only tool called `generate_schema_docs`. This tool consolidates data from PostgreSQL system catalogs into a single, structured response.

**Key capabilities:**
- **Consolidated Schema**: Fetches Tables, Columns, Foreign Keys, RLS Policies, Triggers, and Functions in one request.
- **Dual Format Output**: Returns both a **human-readable Markdown summary** (perfect for generating `DATABASE.md` files) and **structured JSON data** (perfect for AI agent reasoning).
- **Nested Visibility**: In the Markdown output, RLS policies and Triggers are displayed directly under their respective tables for better context.
- **Configurable**: Allows filtering by specific schemas and toggling the inclusion of internal/system functions.

## Describe alternatives you've considered

The current alternative is relying purely on `list_tables (verbose: true)`. However, this captures only table structures and misses:
1.  **RLS Policies**: Crucial for security auditing.
2.  **Triggers**: Essential for understanding side-effects and business logic.
3.  **Custom Functions**: Where complex database-level logic often resides.

Manual JSON-to-Markdown formatting by an LLM is token-expensive and often misses these deeper architectural details.

## Additional context

**Implementation Status:**
I have successfully implemented this tool locally within the `database` feature group. 
- Implemented in PR #278 
- **Tool Name**: `generate_schema_docs`
- **Location**: `packages/mcp-server-supabase/src/tools/database-docs-tools.ts`
- **Tests**: Full test coverage added to `server.test.ts` (including format variants and edge cases). All 172 unit tests are passing.
- **Safety**: This is a strictly read-only operation (`readOnlyHint: true`).

This tool acts as a practical bridge for "Self-Documenting Databases" and empowers developers to instantly generate project-specific schema documentation for onboarding or LLM context initialization.

[...]

PR #278, opened 2026-05-09
feat: add generate_schema_docs tool for database documentation

hasithasandunlakshan:

Feature

Currently, to understand the full structure of a database (tables, columns, RLS policies, triggers, and functions), an AI agent or developer must call multiple tools (e.g., `list_tables` with `verbose: true`, `execute_sql`, etc.) or perform multiple manual queries. There is no single, consolidated way to get a documentation-ready overview of the database schema.

Relevant Issue: #277 

## What is the new behavior?

This PR introduces a new tool, `generate_schema_docs`, which provides a comprehensive, documentation-ready summary of the database schema in a single call.

**Key Features:**
- **Consolidated View**: Fetches tables, columns, foreign keys, RLS policies, triggers, and user-defined functions in one operation.
- **Dual Output Formats**: Supports `markdown` (optimized for humans/AI context windows), `json` (optimized for programmatic use), or `both`.
- **Smart Filtering**: Includes options to filter by specific schemas and toggle the inclusion of internal/system functions.
- **RLS Awareness**: Highlights whether RLS is enabled and lists policies directly under their respective tables in the markdown output.
- **Architectural Alignment**: Correctly integrated into the `database` feature group and wired into the MCP server runtime.

## Additional context

- **Test Coverage**: Added comprehensive unit tests in `server.test.ts` covering various output formats and edge cases. All 172 unit tests are passing.
- **Zod Integration**: Fully typed input and output schemas, allowing for type-safe usage via the AI SDK.
- **Performance**: Uses parameterized SQL queries and grouping logic to minimize token bloat while maintaining high information density.

### DIFF
--- a/packages/mcp-server-supabase/src/pg-meta/functions.sql
+++ b/packages/mcp-server-supabase/src/pg-meta/functions.sql
@@ -1,0 +1,22 @@
+SELECT
+  n.nspname AS schema_name,
+  p.proname AS function_name,
+  pg_get_function_identity_arguments(p.oid) AS identity_arguments,
+  pg_get_function_result(p.oid) AS result_type,
+  l.lanname AS language,
+  CASE p.provolatile
+    WHEN 'i' THEN 'IMMUTABLE'
+    WHEN 's' THEN 'STABLE'
+    ELSE 'VOLATILE'
+  END AS volatility,
+  p.prosecdef AS security_definer,
+  obj_description(p.oid, 'pg_proc') AS comment
+FROM pg_proc p
+JOIN pg_namespace n ON p.pronamespace = n.oid
+JOIN pg_language l ON p.prolang = l.oid
+LEFT JOIN pg_depend d
+  ON d.classid = 'pg_proc'::regclass
+  AND d.objid = p.oid
+  AND d.deptype = 'e'
+WHERE p.prokind = 'f'
+  AND d.objid IS NULL
--- a/packages/mcp-server-supabase/src/pg-meta/index.ts
+++ b/packages/mcp-server-supabase/src/pg-meta/index.ts
@@ -1,7 +1,10 @@
 import { stripIndent } from 'common-tags';
 import columnsSql from './columns.sql';
 import extensionsSql from './extensions.sql';
+import functionsSql from './functions.sql';
+import policiesSql from './policies.sql';
 import tablesSql from './tables.sql';
+import triggersSql from './triggers.sql';
 
 export const SYSTEM_SCHEMAS = [
   'information_schema',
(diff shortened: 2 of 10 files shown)


anp0429:
Nice tool, been trying it out. Ran it against a few schema shapes and hit something on the foreign key output that seemed worth flagging.

For a composite (multi column) foreign key, the FK list looks like it's giving the cartesian product of the source and target columns instead of the actual column pairing.

Quick repro with one composite FK, (a, b) references parent (a, b):
```sql
create table public.parent (a int not null, b int not null, primary key (a, b));
create table public.child (
  a int not null, b int not null,
  constraint child_parent_fk foreign key (a, b) references public.parent (a, b)
);
```
generate_schema_docs comes back with 4 edges for child_parent_fk:
```
public.child.a -> public.parent.a
public.child.a -> public.parent.b   (this pairing doesn't exist)
public.child.b -> public.parent.a   (this one doesn't either)
public.child.b -> public.parent.b
```
Only a->a and b->b are real. The a->b and b->a rows are relationships that aren't actually in the schema. Since the tool's meant for AI reasoning and security auditing, having it report FKs that don't exist felt worth raising.

Tracked it down to the FK subquery in pg-meta/tables.sql — the source and target columns are joined independently:
```sql
... on sa.attrelid = c.conrelid  and sa.attnum = any (c.conkey)
... on ta.attrelid = c.confrelid and ta.attnum = any (c.confkey)
```
any(conkey) against any(confkey) cross-joins every source column with every target column, so an N-column FK gives N² rows. Pairing needs to be positional (conkey[i] with confkey[i]) — unnesting the two arrays together with ordinality keeps them aligned:
```sql
join lateral unnest(c.conkey, c.confkey) with ordinality as cols(conkey, confkey, ord) on true
join pg_attribute sa on sa.attrelid = c.conrelid  and sa.attnum = cols.conkey
join pg_attribute ta on ta.attrelid = c.confrelid and ta.attnum = cols.confkey
```
Same query also backs list_tables (verbose), so this likely affects that path too — this PR just made it visible.

Here's a failing test if it helps, drops in next to the existing schema docs test:
```ts
test('composite foreign key is not a cartesian product', async () => {
  // ...same setup, plus the two tables above...
  const child = result.data.tables.find((t) => t.full_name === 'public.child');
  const pairs = child.foreign_key_constraints.map((f) => `${f.source}=>${f.target}`);
  expect(pairs).not.toContain('public.child.a=>public.parent.b');
  expect(pairs).not.toContain('public.child.b=>public.parent.a');
  expect(child.foreign_key_constraints).toHaveLength(2);
});
```
Also noticed the FK list is the only collection that isn't sorted (tables, policies, triggers, functions all are), so FK order might shift between runs.

Issue #282, opened 2026-05-14
Fix invalid array indexing in tests and correct Edge Function typo


hasithasandunlakshan:
# Bug report

## Describe the bug

The following issues exist in the current test suite:

1. The `delete branch` test case in `server.test.ts` uses an invalid negative array index (`branches[-1]`), which does not work in standard JavaScript environments.
2. There is a typo in one of the Edge Function test descriptions: `eploy edge function validates slug format`.

---

## To Reproduce

1. Run the test suite locally.
2. Navigate to the `delete branch` test case in `server.test.ts`.
3. Observe that `branches[-1]` returns `undefined`.
4. Run Edge Function related tests and notice the typo in the test description.

---

## Expected behavior

1. The test should correctly access the last array element using a valid JavaScript approach.
2. Test descriptions should be typo-free and readable.

---

## Screenshots

Not applicable.

---

## System information

- OS: macOS
- Browser (if applies): N/A
- Version of Node.js: 22.x

---

## Additional context

These issues have been fixed in PR #279.

gregnr:
Fixed in #279

PR #279, opened 2026-05-09
fix(test): resolve edge function test typo and invalid branch array access

hasithasandunlakshan:
## What kind of change does this PR introduce?

Bug fix

## What is the current behavior?

1. The `delete branch` test case in `server.test.ts` fails in standard JavaScript environments due to an invalid negative array index `branches[-1]`.
2. There is a typo in one of the Edge Function test descriptions: `eploy edge function validates slug format`.

## What is the new behavior?

1. Replaces `branches[-1]` with the modern JavaScript `.at(-1)` method to safely access the last element of the array.
2. Corrected the typo from `eploy` to `deploy`.

## Additional context

These issues were identified while implementing new database tools. Resolving them ensures a stable test suite for all contributors.

PR #283, opened 2026-05-14
Hide mutating tools from read-only tools/list

### LINKED ISSUE #281: read-only stdio mode still exposes mutating tools in tools/list
Hi Supabase team,

## Reproduction

```bash
```

Observed tool names include:

```text
apply_migration
deploy_edge_function
create_branch
delete_branch
merge_branch
reset_branch
rebase_branch
```

```json
{
  "name": "apply_migration",
  "annotations": {
    "readOnlyHint": false,
    "destructiveHint": true
  }
}
```

## Expected

## Why this matters

## mcp-lint context

Lubrsy706:
## Summary
- add an optional `listedTools` subset to `createMcpServer` so `tools/list` can differ from callable tools
- use that subset in Supabase read-only mode to hide write tools from MCP clients while preserving existing runtime read-only errors for stale callers
- add a regression test that verifies read-only `tools/list` contains no `readOnlyHint: false` tools

Fixes #281.

## Verification
- pnpm --filter @supabase/mcp-utils typecheck
- pnpm --filter @supabase/mcp-utils build
- pnpm --filter @supabase/mcp-server-supabase typecheck
- CI=1 pnpm --filter @supabase/mcp-server-supabase exec vitest run --project unit src/server.test.ts -t "read-only mode excludes write tools from tools/list"
- CI=1 pnpm --filter @supabase/mcp-server-supabase exec vitest run --project unit src/server.test.ts -t "cannot apply migration in read-only mode"
- pnpm exec biome check packages/mcp-utils/src/server.ts packages/mcp-server-supabase/src/server.ts packages/mcp-server-supabase/src/server.test.ts
- pnpm --filter @supabase/mcp-server-supabase build

Note: the full `src/server.test.ts` suite still hits the existing `delete branch` failure tracked by #282 / #279 on current main.

### DIFF
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -3017,7 +3017,7 @@ describe('tools', () => {
   });
 
   test('read-only mode excludes write tools from tools/list', async () => {
-    const { client } = await setup({
+    const { callTool, client } = await setup({
       readOnly: true,
       features: [
         'docs',
@@ -3046,6 +3046,19 @@ describe('tools', () => {
         .filter((tool) => tool.annotations?.readOnlyHint === false)
         .map((tool) => tool.name)
     ).toEqual([]);
+
+    const result = callTool({
+      name: 'apply_migration',
+      arguments: {
+        project_id: 'test-project-ref',
+        name: 'test-migration',
+        query: 'create table test (id int)',
+      },
+    });
+
+    await expect(result).rejects.toThrow(
+      'Cannot apply migration in read-only mode.'
+    );
   });
 
   test('tool result content is valid JSON', async () => {
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -3016,6 +3016,38 @@ describe('tools', () => {
     }
   });
 
+  test('read-only mode excludes write tools from tools/list', async () => {
+    const { client } = await setup({
+      readOnly: true,
+      features: [
+        'docs',
+        'account',
+        'database',
+        'debugging',
+        'development',
+        'functions',
+        'branching',
+        'storage',
+      ],
+    });
+
+    const { tools } = await client.listTools();
+    const toolNames = tools.map((tool) => tool.name);
+
+    expect(toolNames).toContain('execute_sql');
+    expect(toolNames).not.toContain('apply_migration');
+    expect(toolNames).not.toContain('deploy_edge_function');
+    expect(toolNames).not.toContain('create_branch');
+    expect(toolNames).not.toContain('delete_branch');
+    expect(toolNames).not.toContain('update_storage_config');
+
+    expect(
+      tools
+        .filter((tool) => tool.annotations?.readOnlyHint === false)
+        .map((tool) => tool.name)
+    ).toEqual([]);
+  });
+
   test('tool result content is valid JSON', async () => {
     const org = await createOrganization({
       name: 'My Org',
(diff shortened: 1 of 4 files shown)


packages/mcp-server-supabase/src/server.ts
```diff
@@ -67,6 +67,14 @@ const DEFAULT_FEATURES: FeatureGroup[] = [
 
 export const PLATFORM_INDEPENDENT_FEATURES: FeatureGroup[] = ['docs'];
 
+function removeWriteTools(tools: Record<string, Tool>) {
```
raulb, line 71: This is technically accurate, but could we reuse the existing [`writeToolSet`](https://github.com/supabase/mcp/blob/7e71947/packages/mcp-server-supabase/src/tools/tool-schemas.ts#L136-L143) here instead of a new predicate? 

Same "is it a write tool?" rule now lives in 2 spots (here and in `writeToolSet`) and can drift.

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -3016,6 +3016,38 @@ describe('tools', () => {
...
+    expect(toolNames).not.toContain('create_branch');
+    expect(toolNames).not.toContain('delete_branch');
+    expect(toolNames).not.toContain('update_storage_config');
+
+    expect(
+      tools
+        .filter((tool) => tool.annotations?.readOnlyHint === false)
+        .map((tool) => tool.name)
+    ).toEqual([]);
+  });
```
raulb, line 3062: Thank you for adding tests!

Could we also test that a hidden tool is still *callable*? 

The [`listedTools` doc](https://github.com/supabase/mcp/blob/7e71947/packages/mcp-utils/src/server.ts#L268-L276) promises a runtime error over "unknown tool", but nothing tests it.

raulb (changes requested):
Thank you for the contribution @Lubrsy706 !! (and sorry for late reviewing). I left a couple of suggestions. Hope these made sense! Let me know otherwise.

packages/mcp-server-supabase/src/server.ts
Lubrsy706: Addressed in 09294dc. `server.ts` now filters against the existing `writeToolSet` exported from `tool-schemas.ts`, so `tools/list` and schema filtering share the same write-tool definition. Focused regression test, typecheck, formatting, and all 11 `mcp-utils` tests pass.

packages/mcp-server-supabase/src/server.test.ts
Lubrsy706: Addressed in 09294dc. The read-only `tools/list` regression test now calls the hidden `apply_migration` tool and asserts the specific `Cannot apply migration in read-only mode.` runtime error, rather than `tool not found`. The focused regression test passes.

Lubrsy706:
Implemented both review suggestions in 09294dc, replied to and resolved both threads. The focused regression test, Supabase package typecheck, Biome check, and all 11 `mcp-utils` tests pass. The four GitHub Actions runs currently show `action_required`, so they need maintainer approval before they can execute. @raulb, could you please re-review when convenient?

raulb:
@Lubrsy706 Hey, sorry for the unnecessary hassle! We recently added [this feature](https://github.com/supabase/mcp/pull/334), which would simplify this PR. 

When I was reviewing your PR, I noticed [this other one](https://github.com/supabase/mcp/pull/347) (there were some type errors), which I just fixed. Rather than having you rebase and keep you waiting, I went ahead and made the fix to avoid causing you trouble. I hope to do a better job keeping up with notifications so I don't miss any other contributions you may have. Thanks again, and apologies.

Lubrsy706:
Thanks for taking care of this and for the thoughtful follow-up — no worries at all. Glad the newer implementation covers the fix cleanly.

PR #284, opened 2026-05-14
Add focused parameters to get_logs

### LINKED ISSUE #173: New `get_logs` tool parameters to reduce context bloat
# Feature request

## Is your feature request related to a problem? Please describe.
No

I often debug front/backend problems that require viewing EDGE function logs precisely.  I have a project with more than 30 EDGE functions. They create a lot of logs. The current `get_logs` tool dumps the logs for the past 24 hours. This is great but destroys the fine grained context needed to solve problems at hand. Often I trigger a specific issue and I need cursor to just see the logs for he past 5 minutes.

It would be great if `get_logs` provided additional parameters to help grab the relevant logs from the relevant EDGE function.

## Describe the solution you'd like
`get_logs` fine grained control on time and service logs

> the `get_logs` tool parameters allow LLMs to select time window and sub service such as a specific EDGE function to grab its logs

## Describe alternatives you've considered
Currently I setup a python script that pokes the `get_logs` tool and then filters the output and provides that as context to cursor so it can focus on the issue

One alternative might be to provide a `logs` option in the Supabase CLI. Currently there is not lgos sub-tool within the CLI, but that would be great as well

## Additional context

The get_logs -> Edge functions tool call currently return 40,000 Chars in my project. and the LLM is 100% clueless as to the relevant one because it is so large since there are logs from many unrelated EDGE functions

Lubrsy706:
## Summary
- add optional `minutes`, `limit`, and `search` parameters to the `get_logs` tool
- pass the row limit and search filter into generated log SQL while keeping the existing default 24-hour / 100-row behavior
- add unit coverage for service-specific log search filters and quote escaping

Fixes #173.

## Verification
- pnpm --filter @supabase/mcp-utils build
- pnpm --filter @supabase/mcp-server-supabase typecheck
- CI=1 pnpm --filter @supabase/mcp-server-supabase exec vitest run --project unit src/logs.test.ts
- CI=1 pnpm --filter @supabase/mcp-server-supabase exec vitest run --project unit src/server.test.ts -t "get logs for each service type"
- pnpm exec biome check packages/mcp-server-supabase/src/logs.ts packages/mcp-server-supabase/src/logs.test.ts packages/mcp-server-supabase/src/tools/debugging-tools.ts packages/mcp-server-supabase/src/platform/api-platform.ts packages/mcp-server-supabase/src/platform/types.ts
- pnpm --filter @supabase/mcp-server-supabase build

### DIFF
--- a/packages/mcp-server-supabase/src/logs.test.ts
+++ b/packages/mcp-server-supabase/src/logs.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, test } from 'vitest';
+import { getLogQuery } from './logs.js';
+
+describe('getLogQuery', () => {
+  test('keeps the existing numeric limit overload', () => {
+    const query = getLogQuery('edge-function', 25);
+
+    expect(query).toContain('from function_edge_logs');
+    expect(query).toContain('limit 25');
+  });
+
+  test('adds a service-specific text filter', () => {
+    const query = getLogQuery('edge-function', {
+      limit: 50,
+      search: 'planner',
+    });
+
+    expect(query).toContain('where (');
+    expect(query).toContain("event_message ilike '%planner%'");
+    expect(query).toContain("m.function_id ilike '%planner%'");
+    expect(query).toContain("request.path ilike '%planner%'");
+    expect(query).toContain('limit 50');
+  });
+
+  test('escapes single quotes in text filters', () => {
+    const query = getLogQuery('api', {
+      search: "worker's path",
+    });
+
+    expect(query).toContain("event_message ilike '%worker''s path%'");
+  });
+});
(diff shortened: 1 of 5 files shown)


anp0429:
metadata.status is typed as number (shared-data/log-constants.ts), and elsewhere in the dashboard it's always cast before use, e.g. SAFE_CAST(metadata.status AS INT64) in Logs.constants.ts. In this PR's auth branch it's passed to buildSearchFilter uncast, producing metadata.status ilike '%...%' on a numeric column, while the edge/function_edge branches do cast their status (cast(response.status_code as text)). Should the auth branch cast metadata.status to text for the ilike too, to stay consistent?

Lubrsy706:
Good catch — fixed in 5018cf6 by casting `metadata.status` to text before applying the case-insensitive search filter, matching the existing API/edge-function status handling. I also added a focused regression test for auth status search; all 4 log-query tests pass and Biome is clean. Thanks for flagging this.

PR #290, opened 2026-05-21
Docs/multi project setup

### LINKED ISSUE #233: Unclear how to use the MCP server with multiple projects
# Improve documentation

## Link 

[Supabase MCP server docs](https://supabase.com/docs/guides/getting-started/mcp)

## Describe the problem

If I have multiple Supabase projects, either in the same org or different orgs, it's not clear how we can set up an AI coding assistant to have MCP access to each project, without disabling and re-enabling.

I seem to end up in a loop of OAuth authorisation requests, and confusion as to which MCP server gets used despite telling the agent to use a specifically named agent e.g. use MCP server named `supabase-project1` and verify that it is for a project with id xyz. In my case, I'm using the Codex app.

## Describe the improvement

Perhaps allow a single MCP to be set up for multiple projects with the requirement to confirm the project id (other MCP servers like Sentry seem to do this OK), or provide guidance on the recommended approach for multiple, which perhaps is simply enabling / disabling parallel MCP servers.

AnuragRaut08:
## Summary

Adds documentation explaining how to configure multiple named MCP
server instances for workflows involving multiple Supabase projects.

This includes:

* `.mcp.json` configuration examples
* PAT authentication guidance
* Recommended use of `--read-only` for production
* Explanation of why separate server instances avoid OAuth authorization loops

Fixes #233

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,50 @@ We recommend enabling this setting by default. This prevents write operations on
 `rebase_branch`
 `update_storage_config`.
 
+
+
+
+### Using multiple projects simultaneously
+
+If you work across multiple Supabase projects (in the same org or different orgs), 
+you can run multiple named MCP server instances — one per project — rather than 
+switching a single server between projects.
+
+Create a `.mcp.json` file at your project root:
+
+```json
+{
+  "mcpServers": {
+    "supabase-staging": {
+      "type": "stdio",
+      "command": "npx",
+      "args": ["-y", "@supabase/mcp-server-supabase@latest", "--project-ref=<staging-ref>"],
+      "env": { "SUPABASE_ACCESS_TOKEN": "<your-pat>" }
+    },
+    "supabase-prod": {
+      "type": "stdio",
+      "command": "npx",
+      "args": [
+          "-y",
+          "@supabase/mcp-server-supabase@latest",
+          "--project-ref=<staging-ref>"
+        ],
+      "env": { "SUPABASE_ACCESS_TOKEN": "<your-pat>" }
+    }
+  }
+}
+```
+
+A few things worth knowing:
+
+- **One PAT covers all projects** — Personal Access Tokens are org-scoped, so the same token works for every project in your org.
+- **`--project-ref` locks the server at startup** — each process only ever talks to one project. In clients like Claude Code, tools get namespaced automatically: `mcp__supabase-staging__*` vs `mcp__supabase-prod__*`, so you can tell the agent exactly which server to use.
+- **Use `--read-only` on production** — this restricts to SELECT queries at the protocol level, preventing accidental DDL against prod.
+- **Why not a single server?** — Project scope is resolved at startup. A single server switching projects mid-session must re-authenticate each time, causing the OAuth loop. Two separate processes sidestep this entirely.
+
+Generate a Personal Access Token at `supabase.com/dashboard/account/tokens` and use it as `SUPABASE_ACCESS_TOKEN` to avoid OAuth prompts altogether.
+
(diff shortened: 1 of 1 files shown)


gregnr (commented):
Thanks for the PR @AnuragRaut08! In general we don't recommend connecting Supabase MCP to prod due to the [security risks](https://github.com/supabase-community/supabase-mcp#security-risks) involved, so my concern is that documenting multi-env configurations like this would send mixed messages.

One thing that would be great to add though is a specific callout to prefer project-scoped MCP (e.g. `.mcp.json` files in each repo) over global user-wide MCP (e.g. `~/.claude.json`). This way users can scope each repo to their respective Supabase project and simply switching into each repo directory will automatically inherit the correct MCP configuration. Likely many users already do this, but a specific callout would be helpful.

AnuragRaut08:
Thanks for the feedback! I've updated the guide to recommend project-scoped MCP configuration files (for example, `.mcp.json` within each repository) instead of global user-wide configuration, and clarified the multi-project setup guidance accordingly.

README.md
```diff
@@ -87,6 +87,61 @@ We recommend enabling this setting by default. This prevents write operations on
...
+      "args": [
+        "-y",
+        "@supabase/mcp-server-supabase@latest",
+        "--read-only",
+        "--project-ref=<secondary-ref>"
+      ],
+      "env": {
+        "SUPABASE_ACCESS_TOKEN": "<your-pat>"
+      }
+    }
```
gregnr, line 131: Perhaps we can reduce this to a single entry called `supabase` for demonstration purposes? Also the `stdio` transport is no longer recommended in favour of the remote HTTP transport (`https://mcp.supabase.com/mcp`). Can we please change the config accordingly?

README.md
```diff
@@ -87,6 +87,61 @@ We recommend enabling this setting by default. This prevents write operations on
...
+```
+
+A few things worth knowing:
+
+* **One PAT covers all projects** — Personal Access Tokens are org-scoped, so the same token works for every project in your organization.
+* **`--project-ref` locks the server at startup** — each process only ever talks to one project. In clients like Claude Code, tools get namespaced automatically: `mcp__supabase-staging__*` vs `mcp__supabase-secondary__*`, so you can tell the agent exactly which server to use.
+* **Avoid connecting to production** — prefer development projects whenever possible. If production access is required, use project scoping and read-only mode.
+* **Why not a single server?** — Project scope is resolved at startup. A single server switching projects mid-session may require re-authentication each time. Running separate server instances avoids that workflow entirely.
+
+Generate a Personal Access Token at `supabase.com/dashboard/account/tokens` and use it as `SUPABASE_ACCESS_TOKEN` to avoid OAuth prompts altogether.
```
gregnr, line 143: Remote MCP (http transport) no longer requires a PAT - the happy path is to simply use OAuth to authorize. Perhaps we can omit most or all of these points?

gregnr (commented):
Thanks @AnuragRaut08 🙂 added a few more comments.

AnuragRaut08:
Thanks for the additional feedback @gregnr 

I've updated the documentation to align with the current recommended MCP workflow:

* Replaced the previous stdio transport example with the recommended remote HTTP transport configuration
* Simplified the example to a single `supabase` entry
* Removed the PAT-specific and multi-server workflow details to better reflect the current recommended setup
* Kept the recommendation for repository-scoped `.mcp.json` configuration when working across multiple projects

Pushed the updates for another review. Thanks again for the guidance.

AnuragRaut08:
@gregnr Any follow up?

PR #303, opened 2026-06-09
ci: add force_publish input to release workflow

mattrossman:
Adds a `force_publish` boolean input to the Release workflow's manual dispatch, adapted from the [pattern in Greg's reposh](https://github.com/supabase-community/reposh/blob/faad5c895a472156f54b9b14a3d9623ff8498bb3/.github/workflows/release.yml#L8-L11). Useful when the release workflow fails partway through, [example instance](https://supabase.slack.com/archives/C02AK9166FR/p1780935267920019?thread_ts=1780513141.174739&cid=C02AK9166FR).

When checked, bypasses release-please gating and runs build + all publish steps against current `package.json` versions. Only packages missing from npm actually publish. Also checks MCP registry for existing version before publication to keep the workflow idempotent.

Regenerated mgmt-api types for required check.

Closes AI-828

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,10 +10,6 @@ on:
         description: 'Skip release-please and publish current package.json versions'
         type: boolean
         default: false
-      release_ref:
-        description: 'Required with force_publish. Release tag or commit SHA to publish from.'
-        type: string
-        default: ''
 
 permissions:
   contents: write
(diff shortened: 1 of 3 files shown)


gregnr (approved):

gregnr:
@mattrossman side note - I wonder if we should enforce which ref is used during the manual `force_publish`. My concern is (and this is true for reposh too) that new commits can land in `main` since the failed publish, and if you trigger the manual dispatch on `main` instead of the failed tag, you're publishing new features that weren't meant to land in that version.

mattrossman:
Good point, I'll fold that into this PR to prevent footguns.

mattrossman:
Added a `release_ref` input which accepts a release tag or commit SHA.

Although optional, there's a validation step than enforces it's populated if `force_publish` is enabled, and that it resolves to a tag/SHA.

gregnr:
@mattrossman GitHub Actions already ask for a ref when you perform a manual workflow dispatch:
image

I'm wondering if it would be simpler to just validate the implicit `github.ref` maps to a valid release tag?

mattrossman:
Changed to that approach (except I'm using the split `github.ref_type` / `github.ref_name` for easier validation), the dropdown lets you pick either a branch name or tag so if branch name is picked (or a tag not associated with a GH Release) it fails on validation. It's a little unfortunate it doesn't let you select a commit in the dropdown, only branches or tags. In our monorepo Release Please creates separate tags per package so it's a little arbitrary which package's tag you pick as multiple may point to the same SHA.

PR #323, opened 2026-07-07
fix(server): hide mutating stdio tools in read-only mode fixes #281

### LINKED ISSUE #281: read-only stdio mode still exposes mutating tools in tools/list

ahfoysal:
## Summary
- Ensured `createSupabaseMcpServer` filters tools before registration when `--read-only` is enabled.
- The stdio transport now omits tools with `readOnlyHint: false` in read-only mode, so `tools/list` no longer exposes mutating actions.
- Kept existing behavior intact for adaptive tools (e.g. `execute_sql`) that remain visible with adjusted annotations.

## Verification
- `CI=1 ./node_modules/.bin/vitest run --project integration test/stdio.integration.ts`

Closes #281

### DIFF
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -136,6 +136,14 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
     tools: async () => {
       const contentApiClient = await contentApiClientPromise;
       const tools: Record<string, Tool> = {};
+      const addTools = (next: Record<string, Tool>) =>
+        Object.entries(next).forEach(([name, tool]) => {
+          if (readOnly && tool.annotations?.readOnlyHint === false) {
+            return;
+          }
+
+          tools[name] = tool;
+        });
 
       const {
         account,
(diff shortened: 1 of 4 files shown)


raulb:
Hey! Thank you for your contribution! I was actually reviewing https://github.com/supabase/mcp/pull/283, which seems to fix the same core issue as https://github.com/supabase/mcp/issues/281.

That other PR adds something I particularly like: the fact that write tools are still registered, and it returns a nicer error when in read mode, e.g., "Cannot apply migration in read-only mode." This PR seems to remove that altogether, which I think provides a less-than-ideal UX.

I appreciate your contribution, though, so please keep them coming!

PR #325, opened 2026-07-09
docs: Link directly to docs

raulb:

Updated security best practices link to the Supabase docs instead of using an almost empty section.


barryroodt (approved):
LGTM :shipit:

PR #329, opened 2026-07-14
fix: actionable error for wrong-org permission failures on database tools

barryroodt:
## Problem

Authenticating to the wrong organization (e.g. selecting the wrong org during OAuth login) makes database tools (`list_tables`, `execute_sql`, migrations) fail with a bare:

> You do not have permission to perform this action

Nothing indicates that org selection is the cause — hard to debug for users, and useless to the LLM driving the session.

Linear: [AI-178](https://linear.app/supabase/issue/AI-178/improve-mcp-db-tool-error-for-wrong-organization-selection)

## Change

* New `assertProjectScopedSuccess(response, fallbackMessage, projectId)` in `management-api/index.ts`: intercepts **403 only**, preserves the upstream API message when present, and appends:

  > Access to project '<ref>' was denied. If this project exists, your access token may be scoped to a different organization: re-authenticate with the MCP server and select the organization that owns this project.

  All non-403 errors fall through to the existing `assertSuccess`.
* Wired into the three database platform ops (`executeSql`, `listMigrations`, `applyMigration`), which back `execute_sql`, `list_tables`, `list_extensions`, `list_migrations`, and `apply_migration`.
* Intentionally **not** applied to other endpoints (billing/admin/project ops), where a 403 can mean plan or role restrictions and org-selection advice would mislead. JSDoc on the helper documents this.

Note: the target project's org can't be named in the message — with a wrong-org token the project lookup itself returns 403, so the org is unknowable. The message names the project ref and the likely cause instead.

## Verification

All commands run from `packages/mcp-server-supabase`:

* TDD: started with two failing tool-level regression tests (`execute_sql` and `list_tables` against an MSW 403 override on `/v1/projects/:ref/database/query`), which surfaced the raw upstream message (`You do not have permission to perform this action`) before the fix. Broadened per review to a parametrized suite over all five tools routed through the helper, plus a no-message case that exercises the `fallbackMessage` branch. `pnpm vitest run src/server.test.ts -t "permission denied"`:

  ```
  ✓ tools > permission denied for 'execute_sql' suggests checking organization
  ✓ tools > permission denied for 'list_tables' suggests checking organization
  ✓ tools > permission denied for 'list_extensions' suggests checking organization
  ✓ tools > permission denied for 'list_migrations' suggests checking organization
  ✓ tools > permission denied for 'apply_migration' suggests checking organization
  ✓ tools > permission denied with no upstream message falls back to a generic prefix
  ```
* `pnpm test:unit`: 170/170 passed (10 files)
* `pnpm typecheck` (`tsc --noEmit`): clean
* `pnpm biome check src/management-api/index.ts src/platform/api-platform.ts src/server.test.ts`: clean

## Out of scope

The *timeout* variant from the original user report is gateway-side (tracked as [AI-178](https://linear.app/supabase/issue/AI-178/improve-mcp-db-tool-error-for-wrong-organization-selection) in the Remote MCP (HTTP API) project) and is **not** addressed here.

🤖 AI-assisted; human-reviewed per CONTRIBUTING.md.

### DIFF
--- a/packages/mcp-server-supabase/src/management-api/index.ts
+++ b/packages/mcp-server-supabase/src/management-api/index.ts
@@ -65,3 +65,37 @@ export function assertSuccess<
     throw new Error(fallbackMessage);
   }
 }
+
+/**
+ * Asserts success for project-scoped endpoints, mapping 403 responses to an
+ * actionable message. A permission error on a specific project most commonly
+ * means the access token is scoped to an organization that doesn't own the
+ * project (e.g. the wrong organization was selected during OAuth login).
+ *
+ * Currently applied only to database operations (see AI-178). Before adopting
+ * for other endpoints, check that a 403 there isn't better explained by plan
+ * or role restrictions, where org-selection advice would mislead.
+ */
+export function assertProjectScopedSuccess<
+  T extends Record<string | number, any>,
+  Options,
+  Media extends MediaType,
+>(
+  response: FetchResponse<T, Options, Media>,
+  fallbackMessage: string,
+  projectId: string
+): asserts response is SuccessResponseType<T, Options, Media> {
+  if ('error' in response && response.response.status === 403) {
+    const { data: errorContent } = errorSchema.safeParse(response.error);
+    const apiMessage = (errorContent?.message ?? fallbackMessage).replace(
+      /\.$/,
+      ''
+    );
+
+    throw new Error(
+      `${apiMessage}. Access to project '${projectId}' was denied. If this project exists, your access token may be scoped to a different organization: re-authenticate with the MCP server and select the organization that owns this project.`
+    );
+  }
+
+  assertSuccess(response, fallbackMessage);
+}
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -1344,7 +1344,85 @@ describe('tools', () => {
     await expect(listOrganizationsPromise).rejects.toThrow('Unauthorized.');
   });
 
-  test('permission denied for execute_sql suggests checking organization', async () => {
+  const projectScopedDbTools = [
+    {
+      tool: 'execute_sql',
+      method: 'post',
+      endpoint: '/v1/projects/:projectId/database/query',
+      args: (projectId: string) => ({
+        project_id: projectId,
+        query: 'select 1;',
+      }),
+    },
+    {
+      tool: 'list_tables',
+      method: 'post',
+      endpoint: '/v1/projects/:projectId/database/query',
+      args: (projectId: string) => ({ project_id: projectId }),
+    },
+    {
+      tool: 'list_extensions',
+      method: 'post',
+      endpoint: '/v1/projects/:projectId/database/query',
+      args: (projectId: string) => ({ project_id: projectId }),
+    },
+    {
+      tool: 'list_migrations',
+      method: 'get',
+      endpoint: '/v1/projects/:projectId/database/migrations',
+      args: (projectId: string) => ({ project_id: projectId }),
+    },
+    {
+      tool: 'apply_migration',
+      method: 'post',
+      endpoint: '/v1/projects/:projectId/database/migrations',
+      args: (projectId: string) => ({
+        project_id: projectId,
+        name: 'test-migration',
+        query: 'select 1;',
+      }),
+    },
+  ] as const;
+
+  test.each(projectScopedDbTools)(
+    'permission denied for $tool suggests checking organization',
+    async ({ tool, method, endpoint, args }) => {
+      const { callTool } = await setup();
+
+      const org = await createOrganization({
+        name: 'My Org',
+        plan: 'free',
+        allowed_release_channels: ['ga'],
+      });
+
+      const project = await createProject({
+        name: 'Project 1',
+        region: 'us-east-1',
+        organization_id: org.id,
+      });
+      project.status = 'ACTIVE_HEALTHY';
+
+      mockServer?.use(
+        http[method](`${API_URL}${endpoint}`, () =>
+          HttpResponse.json(
+            { message: 'You do not have permission to perform this action' },
+            { status: 403 }
+          )
+        )
+      );
+
+      const resultPromise = callTool({
+        name: tool,
+        arguments: args(project.id),
+      });
+
+      await expect(resultPromise).rejects.toThrow(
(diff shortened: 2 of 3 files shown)


packages/mcp-server-supabase/src/management-api/index.ts
```diff
@@ -65,3 +65,37 @@ export function assertSuccess<
...
+): asserts response is SuccessResponseType<T, Options, Media> {
+  if ('error' in response && response.response.status === 403) {
+    const { data: errorContent } = errorSchema.safeParse(response.error);
+    const apiMessage = (errorContent?.message ?? fallbackMessage).replace(
+      /\.$/,
+      ''
+    );
+
+    throw new Error(
+      `${apiMessage}. Access to project '${projectId}' was denied. If this project exists, your access token may be scoped to a different organization: re-authenticate with the MCP server and select the organization that owns this project.`
```
mattrossman, line 96: Just noting I see 2 sources of the 403, [one](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/packages/api-core/src/lib/auth/oauth.ts#L48-L49) if token is scoped to wrong org and [other](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/api/apps/mgmt-api/src/guards.ts#L473-L477) if it's missing `database:write` scope. Not sure if 2nd is an issue in practice since we don't (yet) let users pick fine grained scopes, but we may want to handle it in case we support that later.

IIUC, the 403 from missing scopes already [surfaces a helpful error](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/api/apps/mgmt-api/src/guards.ts#L473-L477), so maybe we only want to include the hint about org mismatch when `errorContent?.message` is [missing](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/packages/api-core/src/lib/auth/oauth.ts#L48-L49). Or, make the hint more generic about both of these possibilities:

```suggestion
      `${apiMessage}. Access to project '${projectId}' was denied. If this project exists, your access token may be scoped to a different organization or missing required scopes: re-authenticate with the MCP server, select the organization that owns this project, and approve all requested scopes.`
```

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -1343,6 +1344,81 @@ describe('tools', () => {
...
+      name: 'Project 1',
+      region: 'us-east-1',
+      organization_id: org.id,
+    });
+    project.status = 'ACTIVE_HEALTHY';
+
+    mockServer?.use(
+      http.post(`${API_URL}/v1/projects/:projectId/database/query`, () =>
+        HttpResponse.json(
+          { message: 'You do not have permission to perform this action' },
```
mattrossman: This looks like an error [from our MCP endpoint](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/api/apps/mgmt-api/src/routes/mcp/mcp.controller.ts#L192-L196), whereas I'd expect the error be coming from the `/v1` API in one of the two 403 paths ([1](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/packages/api-core/src/lib/auth/oauth.ts#L48-L49), [2](https://github.com/supabase/platform/blob/6654b435533b6df2fba349d2cabd79e05e520975/api/apps/mgmt-api/src/routes/mcp/mcp.controller.ts#L192-L196)). I only mention it in case we want to test those 2 scenarios to respond differently (wrong org vs missing scope). If we broaden the hint then this doesn't matter though still might be good to align w/ platform message.

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -1343,6 +1344,81 @@ describe('tools', () => {
...
+        query: 'select 1;',
+      },
+    });
+
+    await expect(executeSqlPromise).rejects.toThrow(
+      `You do not have permission to perform this action. Access to project '${project.id}' was denied. If this project exists, your access token may be scoped to a different organization: re-authenticate with the MCP server and select the organization that owns this project.`
+    );
+  });
+
+  test('permission denied for list_tables suggests checking organization', async () => {
```
mattrossman: worth adding a test for `apply_migration` too, since that's also expected to get the hint? or if it gets too repetitive, maybe looping through these in a single test

mattrossman (approved):
Approving since as Cemal mentions [here](https://supabase.slack.com/archives/C051L8U2EJF/p1784047821228399?thread_ts=1784040936.124359&cid=C051L8U2EJF), the insufficient scopes case is probably better handled on the platform side.

packages/mcp-server-supabase/src/management-api/index.ts
barryroodt: Following up here. Good catch on the two distinct 403 sources.

**We're deliberately leaving the missing-scope case to the platform for now.** Per your approval note (and Cemal's thread), the insufficient-scopes 403 reads better surfaced platform-side, and since we don't expose fine-grained scopes yet there's no missing-scope 403 for a user to actually hit today.

**The upstream message is preserved either way**, so if a scope 403 does carry a helpful message, it still shows up as the prefix and we only append the org hint after it. When fine-grained scopes land, this is exactly the spot to revisit, either the `errorContent?.message`-present gate you suggested or the broadened wording.

Very open to broadening the copy now if you'd rather get ahead of that; it just felt like guessing at a case users can't trigger yet, so we wanted it to be a deliberate call rather than baked in early. 🙂

packages/mcp-server-supabase/src/server.test.ts
barryroodt: Nice, this pushed me to make the mock more honest.

**Added a dedicated no-message test** that returns a 403 with no `message` field (a `{}` body), which trips the `errorSchema` parse and exercises the `fallbackMessage` branch (previously untested), asserting the generic `Failed to execute SQL query.` prefix rather than the MCP-endpoint string. That lines up with the missing-message shape you flagged on the wrong-org path.

**Kept the populated-message cases too**, since a 403 that does carry an upstream message (like the mgmt-api guard) should still preserve it. So both shapes are covered now, in 04eee86.

packages/mcp-server-supabase/src/server.test.ts
barryroodt: Done, and went with the loop since you're right that it got repetitive.

**Parametrized over every tool routed through `assertProjectScopedSuccess`**: `execute_sql`, `list_tables`, `list_extensions`, `list_migrations`, and `apply_migration`. So both migration paths are now covered alongside the query ones, in 04eee86.

PR #333, opened 2026-07-16
feat!: add query_logs tool for custom log queries

jordienr:
## What

Adds a new `query_logs` tool to the `debugging` feature group. It runs a custom **ClickHouse SQL query** against a project's unified logs stream, for cases where the `get_logs` service presets are too coarse (filtering, aggregating, or joining across log fields).

Tracks **AI-701**. Builds on the ClickHouse logs endpoint work in O11Y-1813 and the `get_logs` ClickHouse migration (#326).

## Breaking change

This ships as `feat!:` per team alignment (see PR discussion): on platforms that implement ClickHouse-backed querying (hosted/production), `query_logs` is the tool discoverable via `tools/list`, and `get_logs` is hidden from discovery — though it remains callable via `tools/call` for any client still holding an older tool list. On platforms without ClickHouse support (CLI/self-hosted), the reverse holds: `get_logs` stays listed and `query_logs` is not registered at all.

`DebuggingOperations.queryLogs` is optional on the `SupabasePlatform` type (published via the `./platform` subpath export). Implementers outside this repo (CLI, studio, the hosted controller) are unaffected if they don't implement it — they simply keep `get_logs` as their only logs tool.

## How it addresses the observability team's concerns

1. **Prompt injection / security** — log content is user-controllable (same class of risk as `execute_sql`), so the result is wrapped in `wrapWithUntrustedDataBoundary`, the same best-effort guardrail `execute_sql` uses. Not foolproof, but consistent with the existing arbitrary-query tool.
2. **Cost / scalability** — starts conservative: defaults to a 24h window, and both the client (`resolveLogWindow`) and the analytics endpoint cap the requested range at 24h. No unbounded queries.
3. **ClickHouse, not BigQuery** — hits `GET /v1/projects/{ref}/analytics/endpoints/logs` (ClickHouse / logs.all.otel) and takes raw ClickHouse-dialect SQL. Does not touch the deprecated BigQuery-backed `logs.all`.
4. **No POST on v1** — uses `GET` with `sql` as a query param. No new POST handler.
5. **CLI / self-hosted** — `query_logs` is only registered when the platform implements `queryLogs`; CLI/self-hosted platforms don't, so they keep `get_logs` as their only logs tool and never see `query_logs` at all (see Breaking change above).

## Details

- `query_logs` params: `project_id`, `sql`, optional `iso_timestamp_start`/`iso_timestamp_end`.
- Timestamps are validated as ISO 8601 with an explicit UTC `Z` suffix or offset (`z.iso.datetime({ offset: true })`), so the constraint is visible in the tool's JSON schema and offset-less (ambiguous local-time) timestamps are rejected up front.
- `resolveLogWindow` (shared by `get_logs` and `query_logs`) anchors the default start to 24h before the resolved end, normalizes accepted timestamps to canonical UTC ISO strings, and rejects an inverted/equal window or one exceeding 24h — unit-tested directly in `debugging-tools.test.ts`.
- Read-only enforcement of the SQL itself is left to the backend; no client-side SQL parsing.
- Tool descriptions carry no environment-specific wording (hosted/self-hosted/CLI) — routing is handled structurally via registration/hiding, not prose the model has to reason about.

## Verification

- `tsc --noEmit` clean, `biome ci` clean.
- All unit + integration suites pass (incl. `logs.test.ts`, `server.test.ts`, `debugging-tools.test.ts`). The 5 failing `|e2e|` checks need live `SUPABASE_ACCESS_TOKEN` / `ANTHROPIC_API_KEY`, which this fork PR doesn't receive — confirmed passing on an identical-commit mirror pushed directly to the repo (#341).
- Manually verified `query_logs` end-to-end against a real staging project (ClickHouse query executed, results returned, untrusted-data wrapping applied).
- Eval coverage: supabase/evals#100 (draft, blocked on this shipping in a released `@supabase/mcp-server-supabase` version) verifies the agent successfully uses `query_logs` on a hosted-shaped fixture where it's the only logs tool listed.

## Post-merge follow-ups

Per Rodriguespn — this adds a tool and changes `get_logs`'s discoverability, so per [CONTRIBUTING's breaking-change guidance](https://github.com/supabase/mcp/blob/main/CONTRIBUTING.md#breaking-changes):

- [ ] Release a new `@supabase/mcp-server-supabase` version (automatic via release-please once merged as `feat!:`)
- [ ] Update the tool list at [supabase.com/mcp](https://supabase.com/mcp#available-tools)
- [ ] Bump `MCP_SERVER_VERSION` in `supabase/evals` and un-draft evals#100
- [ ] Open a PR on `platform` to add `query_logs` to the remote MCP server
- [ ] Update the `@supabase/mcp-server-supabase` version pin inside `studio`
- [ ] Resubmit the ChatGPT app to pick up the new tool and reworded `get_logs` description (frozen fields)

> [!NOTE]
> Per CONTRIBUTING, feature PRs should track an accepted issue — this tracks AI-701.


barryroodt (commented):
Really clean addition, Jordi. It mirrors the merged `get_logs` sibling almost exactly (same endpoint, same untrusted-data wrapping, same 24h defaulting), and I reproduced the checks against the PR head: `tsc --noEmit`, `biome ci .`, and the `server.test.ts` + `logs.test.ts` suites (111 tests) all pass. Nothing here blocks merge. A couple of small things worth a look:

**The `sql` description is missing `function_edge_logs`.** The source list guides the model on what to filter by, and right now it lists `function_logs` (edge-function runtime console output) but not `function_edge_logs` (the invocation/request logs the `get_logs` `edge-function` preset queries, see `logs.ts:37`). Since that description is effectively the contract the model writes SQL against, a model following it can't reach invocation logs and will filter on the wrong source. Everything else in the list looks right.

**`query_logs` has no execution test yet.** The test change adds it to the tool-listing assertion, which is exactly right for registration, but the new timestamp-defaulting / passthrough / wrapping logic in `execute` is the bug-prone part and it's currently untested (whereas `get_logs` has three). `server.test.ts:1567` is a ready-made template and the existing `/endpoints/logs` mock already covers it, so no new mock needed. Happy to leave this as a fast-follow if you'd rather not expand scope here.

**Tiny nit:** `sql: z.string()` could take `.min(1)` to match `executeSqlOptionsSchema.query`, so an empty query gives a clear validation message instead of an opaque backend error. Very much a nice-to-have.

None of this is load-bearing for shipping. The `function_edge_logs` line is the one I'd genuinely want fixed before merge; the rest are optional.

jordienr:
Thanks for the thorough review! Addressed all three in f37cfc5:

- **`function_edge_logs` missing from the `sql` description** — added it to the source-hint list, so a model can now reach edge function invocation logs (not just `function_logs` runtime output).
- **No execution test for `query_logs`** — added three, mirroring the `get_logs` template at `server.test.ts`: sql passthrough + timestamp defaulting, custom-window forwarding, and empty-query rejection.
- **`sql` nit** — added `.min(1)` on both the tool schema and `queryLogsOptionsSchema`, matching `executeSqlOptionsSchema.query`.

`tsc --noEmit`, `biome ci .`, and the `server.test.ts` + `logs.test.ts` suites (114 tests) all pass.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

barryroodt (commented):
Thanks for the quick turnaround @jordienr  

- `function_edge_logs` added to the source list
- `.min(1)` on the schema
- behavioral tests for `query_logs`

LGTM

@mattrossman, @Rodriguespn - since I'm still the new guy, perhaps a quick scan and thumbs-up from either of you would be advisable 😁

packages/mcp-server-supabase/src/platform/api-platform.ts
```diff
@@ -277,6 +279,30 @@ export function createSupabaseApiPlatform(
 
       return response.data;
     },
+    async queryLogs(projectId: string, options: QueryLogsOptions) {
```
Ziinc, line 282: ~extending the getLogs to accept SQL would be a simpler option, so that it accepts either `service`~

Ziinc (changes requested):

Ziinc:
on further thought, let's leave both tools and potentially deprecate and remove getLogs in favour of queryLogs.
Note that Clickhouse backed querying is only available for production, cli/self-hosted will not have it yet and we need to document that discrepancy as this will break if people try querying against cli. they should use the getLogs tool until full support is provided.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -35,6 +35,32 @@ const getLogsOutputSchema = z.object({
   result: z.unknown(),
 });
 
+const queryLogsInputSchema = z.object({
+  project_id: z.string(),
+  sql: z
+    .string()
+    .min(1)
+    .describe(
+      "A read-only ClickHouse SQL query to run against the project's unified logs stream. Logs are exposed through a `logs` table; filter by `source` (e.g. 'edge_logs', 'postgres_logs', 'function_edge_logs', 'function_logs', 'auth_logs', 'storage_logs', 'realtime_logs', 'workflow_run_logs') and read nested fields via `log_attributes['<key>']`."
```
Ziinc, line 44: would it be better to link documentation to let the llm self-discover?

Ziinc (approved):

packages/mcp-server-supabase/src/platform/api-platform.ts
jordienr: Kept `query_logs` as a separate tool per your follow-up ("deprecate and gradually remove getLogs in favour of queryLogs"). `get_logs` is now marked deprecated on hosted projects in favour of `query_logs`, so the two stay distinct rather than one overloaded `service | sql | lql` param. LQL isn't wired up here — happy to add it as a follow-up if you want a second query mode.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: Kept the concise inline source hints for now — they're the offline contract the model writes SQL against and don't risk link rot. Happy to swap in a docs link if there's a canonical page for the ClickHouse logs schema/sources; point me at it and I'll reference it (or drop the inline list in favour of it).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

jordienr:
Addressed the deprecation + platform-availability guidance in a4d7228:

- `get_logs` is now marked **deprecated on hosted (production) projects in favour of `query_logs`**, while remaining the documented path for local (CLI) and self-hosted projects.
- `query_logs` description now states it's **hosted-only** — ClickHouse-backed querying isn't available on CLI/self-hosted yet, and to use `get_logs` there — so a model won't try it against an unsupported project and break.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

barryroodt:
Following up with fresh numbers against the current head (ecc456d)

Rerun of our eval A/B against the new head: passed all 3 eval checks on the first attempt, and the a4d7228 deprecation note fully flipped tool selection, 8 `query_logs` calls and zero `get_logs` (previously the model mixed both). The first genuine ClickHouse aggregation `(countIf(toInt32OrZero(...)))` succeeded as-is. For contrast, the `get_logs`-only baseline on `main` had one run die on context length counting raw preset rows, so the deprecation direction looks well supported.

One finding: the 24 hour default still silently widens narrow questions. The eval asks about the last 15 minutes; all 8 `query_logs` calls in the fresh run passed no timestamps, inheriting the 24 hour window, even though the description now mentions `iso_timestamp_start`/`iso_timestamp_end`. Mentioning the window permissively isn't landing; in production that means wrong counts whenever older logs exist.

Suggestion: same mechanism as a4d7228 -  make the line directive, something like "when the user asks about a specific time range, always pass iso_timestamp_start/iso_timestamp_end to match it". 

Unrelated side effect: this exercise surfaced drift in our own evals fixture, fixed in supabase/evals#99.

jordienr:
Thanks for the fresh eval run — great to see the deprecation note flip selection cleanly, and good catch on the time-range default.

Fixed in ea6d163: the timestamp guidance is now a directive rather than a permissive mention, using the same mechanism as the deprecation line. Both `query_logs` and `get_logs` now say:

> When the user asks about a specific time range, always pass iso_timestamp_start and iso_timestamp_end to match it; otherwise the query defaults to the last 24 hours and will return results from a wider window than intended.

Applied to `get_logs` too since it shares the identical 24h defaulting. Whenever you next rerun the A/B I'd be curious whether the 15-minute question now passes an explicit window.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

barryroodt:
Nice! Reran the A/B against ea6d163 and the directive line does exactly what we hoped: the model now passes an explicit 15 minute window matching the question on every analytical `query_logs` call (4 of 4), and the only windowless calls are orientation probes (a `now()` check and a quick sample of the data), which is the behaviour you'd want. Still 3/3 checks on the first attempt, still 100% `query_logs` over `get_logs` on hosted.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -35,6 +35,32 @@ const getLogsOutputSchema = z.object({
...
+    ),
+  iso_timestamp_start: z
+    .string()
+    .optional()
+    .describe(
```
Rodriguespn, line 49: nit: would be great to have the default value in the description as well. I'm assuming that the default here is now() but it's always good to be explicit with LLMs

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -35,6 +35,32 @@ const getLogsOutputSchema = z.object({
...
+      'The start of the log window as an ISO 8601 timestamp. The API caps the requested range at 24 hours.'
+    ),
+  iso_timestamp_end: z
+    .string()
+    .optional()
+    .describe(
```
Rodriguespn, line 55: nit: Same as above. Here I assume the default is now() + 24h (?)

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: Good call. Made both defaults explicit on the `iso_timestamp_start`/`iso_timestamp_end` descriptions (both `query_logs` and `get_logs`): end defaults to the current time, start to 24 hours before the end. Pushed in the latest commit.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: Already covered by the same commit (b36a754) — this line now reads "Defaults to the current time." Small correction on the assumption: the end defaults to now() (not now()+24h); the window extends *backward* from the end, so start defaults to 24h before it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -49,7 +75,7 @@ const getAdvisorsOutputSchema = z.object({
...
+      'Gets logs for a Supabase project by service type. When the user asks about a specific time range, always pass iso_timestamp_start and iso_timestamp_end to match it; otherwise each call defaults to the last 24 hours and will return logs from a wider window than intended. The window can be up to 24 hours. Edge Function logs are split by kind: `edge-function` returns invocation/request logs, while `edge-function-runtime` returns console output from inside the function. Query one service first, then correlate with other services by timestamp or error anchors. Do not poll get_logs in a loop. Deprecated on hosted (production) projects in favour of `query_logs`, which supports custom ClickHouse queries; prefer `query_logs` there. Continue using `get_logs` on local (CLI) and self-hosted projects until ClickHouse-backed querying is available for them.',
```
Rodriguespn: I'm curious if the agent will correctly call `query_logs` when working with the hosted version and call `get_logs` when working with self-hosted. Imo, we should have an eval scenario or test it locally as a sanity check for these two cases. 

It can be as simple as:

Prompt 1: "I have a project hosted on Supabase and what to get the logs from it. From the tools you have available from the Supabase MCP, which tool would you call? Don't call it, just name it."

Prompt 2: It's the same but for self-hosted projects

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -60,6 +86,19 @@ export const debuggingToolDefs = {
       openWorldHint: false,
     },
   },
+  query_logs: {
+    description:
+      "Runs a custom read-only ClickHouse SQL query against a Supabase project's unified logs stream. Use this when `get_logs` (service presets) is too coarse and you need to filter, aggregate, or join across log fields. When the user asks about a specific time range, always pass iso_timestamp_start and iso_timestamp_end to match it; otherwise the query defaults to the last 24 hours and will return results from a wider window than intended. The window can be up to 24 hours. Do not poll this tool in a loop. Only available on hosted (production) Supabase projects; ClickHouse-backed querying is not yet available for local (CLI) or self-hosted projects, where you should use `get_logs` instead.",
```
Rodriguespn: I find "`Use this when `get_logs` (service presets) is too coarse and you need to filter, aggregate, or join across log fields`" and "`ClickHouse-backed querying is not yet available for local (CLI) or self-hosted projects, where you should use `get_logs` instead."` confusing. 

The first sentence suggests to use this tool when `get_logs` is not enough but the last sentence indicates that `query_logs` is not always available. 

What should the model do when `get_logs` is not enough and we don't have `query_logs` available?

jordienr: fixed in d9ddd0b

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: Added eval scenarios in supabase/evals#100 (hosted → `query_logs`, self-hosted → `get_logs`, deterministic on tool-call endpoint). Draft there until `query_logs` ships in a released mcp-server version and `MCP_SERVER_VERSION` is bumped.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
barryroodt: Looks like this landed already, the description now states the default explicitly ("24 hours before the end of the window"). One wrinkle with how the code delivers on that promise, picking it up in the thread below so it stays in one place.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
barryroodt: The explicit defaults are in now (end: current time), so the nit itself is resolved. But checking the code against the new wording surfaced a real one: **the description now promises end-anchored defaulting, the code does clock-anchored.** In the execute handler, `startTimestamp` is always computed as now() minus 24h, even when you supply `iso_timestamp_end`. So `iso_timestamp_end: <yesterday 09:00>` with no start gives the window [now minus 24h, yesterday 09:00], inverted or empty, while the description says start defaults to 24 hours before *the end*.

Fix is small: derive the end first (supplied value or now), then default start from that end, or require both-or-neither in the schema. The start-only case is worth a glance too ([start, now] can exceed the 24h cap; presumably the API rejects, but the error will read as mysterious to a model).

packages/mcp-server-supabase/src/tools/debugging-tools.ts
barryroodt: We have half of this measured already, behaviourally rather than by asking: on our hosted-shaped eval fixture, after the deprecation note landed, claude-sonnet-5 organically selected `query_logs` over `get_logs` 8 calls to 0 and 7 to 0 across fresh runs of an aggregation task (it previously mixed both). So the hosted arm looks solid.

One scoping note on those numbers (thanks to @mattrossman's clarification on Slack): our harness drives the legacy stdio server, so it measures how the shared tool descriptions steer the model, not deployment wiring. Production hosted and CLI are separate `SupabasePlatform` implementations, which changes what the self-hosted arm of your test should look like: rather than both environments presenting identical tool lists, the clean design is per-function optionality (extending the #103 feature-group pattern) so the CLI implementation simply never exposes `queryLogs`. At that point the self-hosted selection question mostly answers itself, models can't call what isn't listed.

Your name-the-tool prompt is still a cheap sanity check of the "use `get_logs` there" description line in the meantime. Once a branch exists with the per-function gating, we'll run the proper two-arm behavioural version in the eval harness.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
barryroodt: Sharpest catch in the review, and we can answer "what should the model do" with data instead of a guess: in our A/B baseline (main, `get_logs` only), the model facing an aggregation question fell back to counting raw preset rows client-side, and one run died on context length before answering. That is what currently lives in the gap you found.

Three ways out, in increasing order of rightness:
1. state the fallback explicitly in the description ("on CLI/self-hosted use `get_logs` and process results client-side")
2. gate registration so `query_logs` simply isn't listed where it can't work (the temporary disable from the Slack thread, which makes the contradiction invisible)
3. longer-term, have the API declare its dialect

We'd vote (2) now, (3) later; (1) only if (2) doesn't make this PR. 🙂

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: fixed in c1bdc7f — the handler now derives the end first (supplied or now), then defaults start to end − 24h, so an end-only window is no longer inverted. Shared by get_logs and query_logs, with a regression test. Left the start-only >24h case to the API cap as before (documented in the param description).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: fixed in c1bdc7f (the end-anchoring wrinkle from the sibling thread).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: Agreed. supabase/evals#100 adds exactly these two arms (hosted → query_logs, self-hosted → get_logs) and the PR body calls out the same caveat: with identical tool lists it measures description steering, not real environment awareness. It's a draft pending a released query_logs + MCP_SERVER_VERSION bump, and if we go with capability-gated registration the self-hosted arm becomes 'not listed' rather than 'not chosen'.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: Great data — thank you. Decision from the PR author: ship option (1) in this PR and track option (2) as a fast-follow.

- **(1) is already in place**: both descriptions state the fallback — query_logs says use get_logs on CLI/self-hosted, and get_logs is documented as the only logs tool there. So the model has explicit guidance for the gap you found rather than silently counting raw rows.
- **(2) capability-gated registration** is the right end state and needs the capability-flag branch you mentioned (the server can't distinguish environments today). Tracking it as a fast-follow; when it lands, query_logs simply won't be listed on CLI/self-hosted and evals#100 switches from 'not chosen' to 'not listed'.
- **(3)** API-declared dialect: longer-term, as you noted.

Not blocking this PR on (2).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/mcp-server-supabase/src/platform/types.ts
```diff
@@ -212,6 +219,7 @@ export type EdgeFunctionsOperations = {
 
 export type DebuggingOperations = {
   getLogs(projectId: string, options: GetLogsOptions): Promise<unknown>;
+  queryLogs(projectId: string, options: QueryLogsOptions): Promise<unknown>;
```
Rodriguespn: Making `queryLogs` a required in `DebuggingOperations` breaks every external `SupabasePlatform` implementer on upgrade (this type ships via the `./platform` subpath export and the CLI/studio and mcp.supabase.com controller implement it out-of-repo), and a stale JS implementer would still pass the `if (debugging)` group check, so `query_logs` gets listed in `tools/list` and then crashes with `TypeError: debugging.queryLogs is not a function` at call time. 

We agreed [here](https://supabase.slack.com/archives/C08N7894QTG/p1784626393059569?thread_ts=1784624411.491679&cid=C08N7894QTG) that this tool should be platform-API-only, and [Matt proposed exactly this mechanism](https://supabase.slack.com/archives/C08N7894QTG/p1784642304796489?thread_ts=1784624411.491679&cid=C08N7894QTG): make it `queryLogs?(…)` and register the tool only when the platform provides it. Then self-hosted/CLI need no changes and the tool is simply absent instead of erroring. 


This is also a breaking change to a published type, so can you please change the title of this PR to include `feat!:` and also make sure that the merged commit that lands on main starts with `feat!:` please? 
Check [CONTRIBUTING's expand/contract rule](https://github.com/supabase/mcp/blob/main/CONTRIBUTING.md#breaking-changes) for more details.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -91,16 +159,25 @@ export function getDebuggingTools({
...
           service,
-          iso_timestamp_start:
-            iso_timestamp_start ?? startTimestamp.toISOString(),
-          iso_timestamp_end: iso_timestamp_end ?? endTimestamp.toISOString(),
+          ...resolveLogWindow(iso_timestamp_start, iso_timestamp_end),
+        });
+        return { result: wrapWithUntrustedDataBoundary(result) };
+      },
+    }),
+    query_logs: injectableTool({
```
Rodriguespn: Related to the `types.ts` [comment](https://github.com/supabase/mcp/pull/333/changes#r3713331390): this registration is unconditional, so every platform with a `debugging` group advertises `query_logs`, including BigQuery-dialect (CLI/self-hosted) backends where the description itself says the query will fail. 

Guarding with `if (debugging.queryLogs)` (once the member is optional) hides the tool cleanly on unsupported platforms.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -75,6 +114,35 @@ export const debuggingToolDefs = {
...
+    throw new Error(
+      `Invalid iso_timestamp_start: "${start}". Expected an ISO 8601 timestamp.`
+    );
+  }
+
+  if (startMs >= endMs) {
+    throw new Error('iso_timestamp_start must be before iso_timestamp_end.');
+  }
+
+  return { iso_timestamp_start: start, iso_timestamp_end: end };
```
Rodriguespn: `resolveLogWindow` validates with `Date.parse` but forwards the **original strings** verbatim. 

Two consequences:  (a) [`Date.parse` accepts non-ISO strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) like `"Feb 1 2024"` even though the error message promises ISO 8601 rejection and (b) offset-less timestamps are parsed as server-local time while the management API treats them as UTC. 

I propose normalizing both with `new Date(ms).toISOString()` before returning fixes both, or move format enforcement into the schema with `z.iso.datetime()` so the constraint also shows up in the JSON Schema the model sees.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -91,16 +159,25 @@ export function getDebuggingTools({
         iso_timestamp_start,
         iso_timestamp_end,
       }) => {
-        const endTimestamp = new Date();
-        const startTimestamp = new Date(
-          endTimestamp.getTime() - 24 * 60 * 60 * 1000
-        ); // Last 24 hours
-
         const result = await debugging.getLogs(project_id, {
```
Rodriguespn, line 173: Just a heads-up that routing `get_logs` through `resolveLogWindow` silently changes existing behavior: inputs that previously flowed through to the API are now rejected client-side, and a supplied `iso_timestamp_end` now anchors the default start instead of using `now-24h`. 

That's arguably all improvement, but none of the six new validation tests cover `get_logs`, mirroring the inverted/malformed/anchoring cases for `get_logs` (or exporting `resolveLogWindow` and unit-testing it once) would pin the shared behavior.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -75,11 +114,41 @@ export const debuggingToolDefs = {
...
+
+  const start = iso_timestamp_start ?? new Date(endMs - DAY_MS).toISOString();
+  const startMs = Date.parse(start);
+  if (Number.isNaN(startMs)) {
+    throw new Error(
+      `Invalid iso_timestamp_start: "${start}". Expected an ISO 8601 timestamp.`
+    );
+  }
+
+  if (startMs >= endMs) {
```
Rodriguespn, line 139: nit: Both tool descriptions say "The API caps the requested range at 24 hours", but nothing pre-validates. 

Cheap win: `if (endMs - startMs > DAY_MS) throw new Error('The log window can be at most 24 hours.')` plus the matching unit test.

packages/mcp-server-supabase/src/platform/api-platform.ts
Rodriguespn, line 282: q: Does [`/v1/.../analytics/endpoints/logs`](https://supabase.com/docs/reference/api/v1-get-project-logs) hard-rejects non-SELECT statements?

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -2072,6 +2072,239 @@ describe('tools', () => {
...
+        arguments: {
+          project_id: project.id,
+          sql: 'select id from logs limit 1',
+          iso_timestamp_end: 'not-a-timestamp',
+        },
+      })
+    ).rejects.toThrow();
+  });
+
+  test('query logs rejects a start at or after the end', async () => {
```
Rodriguespn, line 2260: nit: Adding the equal-timestamps case, and giving these negative tests a message matcher (`rejects.toThrow(/must be before/)`) instead of bare `rejects.toThrow()`, would make them mean what they say. 

Same theme at line 2075: the default-window assertions only check the params are truthy, an exact assertion would catch a wrong default or swapped bounds.

Rodriguespn:
Since this adds a tool and rewords the `get_logs` description, two post-merge follow-ups from [CONTRIBUTING](https://github.com/supabase/mcp/blob/main/CONTRIBUTING.md#breaking-changes) apply and are worth tracking in the PR description: 

1. Updating the tool list at [Supabase MCP docs](https://supabase.com/docs/guides/ai-tools/mcp#debugging)
2. A ChatGPT app resubmission (tool list and descriptions are frozen fields). 

Also linking the eval evidence here: the [pkg.pr.new preview build](https://github.com/supabase/mcp/pull/333#issuecomment-5004663705) of this branch passed the full supabase/evals regression suite (8/8, identical to the published-version baseline - [CI run](https://github.com/supabase/evals/actions/runs/30908932414)) and locally against the three debugging evals that exercise the analytics logs endpoint ([evals#79](https://github.com/supabase/evals/pull/79), with transcripts showing agents successfully using `query_logs` with `source = 'function_edge_logs'` ClickHouse SQL.

Conditions of the run: model `claude-sonnet-5` (reasoning effort `high`)

Thanks for taking care of this @jordienr!

Rodriguespn (approved):
LGTM, thanks for taking another look @jordienr. Could you please address the nit comments I left, whether implementing them or closing them. Up to you 🙁 


> [!IMPORTANT]  
> One last thing before merging, this is also a breaking change as we changed `get_logs` description and `inputSchema`. 
> **Can you change the title of this PR to include `feat!:` and also make sure that the merged commit that lands on main starts with `feat!:` please?** 
> 
> Check [CONTRIBUTING's expand/contract rule](https://github.com/supabase/mcp/blob/main/CONTRIBUTING.md#breaking-changes) for more details.

Rodriguespn (changes requested):
Please check [this slack message](https://supabase.slack.com/archives/C051L8U2EJF/p1785933371166709?thread_ts=1785774285.474099&cid=C051L8U2EJF) and the previous ones in that thread for more context on the step back

packages/mcp-server-supabase/src/platform/types.ts
jordienr: fixed in ab36b09

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: fixed in ab36b09

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: fixed in a8e5a98

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: fixed in a8e5a98

packages/mcp-server-supabase/src/tools/debugging-tools.ts
jordienr: fixed in a8e5a98

packages/mcp-server-supabase/src/server.test.ts
jordienr: fixed in 286e9dc

packages/mcp-server-supabase/src/platform/api-platform.ts
jordienr: Asked the o11y team here: https://supabase.slack.com/archives/C08N7894QTG/p1786096927989369

Rodriguespn (approved):
Ran the CI locally and it's green. In general LGTM, thanks for the patience @jordienr. I'm approving this PR to unblock you, but wanted to ask if we can consider implementing the `logsDialect` + dynamic description approach that [Matt suggested](https://supabase.slack.com/archives/C08N7894QTG/p1784660787397029?thread_ts=1784624411.491679&cid=C08N7894QTG).

Seems like a low-hanging fruit to roll out `query_logs` across all envs without having to juggle multiple conditions depending on the environment. We're **not** suggesting implementing ClickHouse for logs on self-hosted/CLI for this — self-hosted/CLI would keep using its existing BigQuery-dialect logs endpoint ([Logflare `logs.all`](https://github.com/supabase/supabase/blob/master/apps/studio/lib/api/self-hosted/mcp.ts#L120-L138)). The idea is just to **hide `get_logs` on all envs** and **change `query_logs`'s description/behavior depending on the env's dialect**, using the strategy Matt described.

Essentially, what I'm suggesting is the following:

**1. Add a `logsDialect` to `DebuggingOperations`** so each platform can declare which SQL its logs endpoint speaks — [`platform/types.ts#L220-L223`](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/platform/types.ts#L220-L223):

```ts
export type DebuggingOperations = {
  logsDialect?: 'clickhouse' | 'bigquery'; // <- new
  getLogs(projectId: string, options: GetLogsOptions): Promise<unknown>;
  queryLogs?(projectId: string, options: QueryLogsOptions): Promise<unknown>;
  ...
};
```

**2. Each platform hardcodes its dialect and implements `queryLogs`.** Hosted/platform-API already implements `queryLogs` (ClickHouse) — just add the dialect at [`platform/api-platform.ts#L255`](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/platform/api-platform.ts#L255):

```ts
const debugging: DebuggingOperations = {
  logsDialect: 'clickhouse', // <- new
  ...
};
```

Self-hosted/CLI sets `logsDialect: 'bigquery'` and implements `queryLogs` by passing the model's SQL straight through to the same BigQuery-backed analytics endpoint `getLogs` already uses (the endpoint takes an arbitrary `sql` param), in [`supabase/apps/studio/lib/api/self-hosted/mcp.ts` `getDebuggingOperations`](https://github.com/supabase/supabase/blob/master/apps/studio/lib/api/self-hosted/mcp.ts#L116-L138):

```ts
return {
  logsDialect: 'bigquery', // <- new
  async getLogs(projectRef, options) { /* unchanged: getLogQuery(service) */ },
  async queryLogs(projectRef, options) { // <- new, mirrors getLogs
    const { data, error } = await retrieveAnalyticsData({
      name: 'logs.all',
      projectRef,
      params: {
        sql: options.sql, // pass the model's SQL instead of a preset
        iso_timestamp_start: options.iso_timestamp_start,
        iso_timestamp_end: options.iso_timestamp_end,
      },
    })
    if (error) throw error
    return data
  },
  ...
}
```

**3. `query_logs`'s description switches on the dialect** — no string juggling, just pick the whole description. When the dialect is `bigquery`, reuse the existing [`get_logs` description](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/tools/debugging-tools.ts#L76-L78); when it's `clickhouse`, use [the one this PR already writes for `query_logs`](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/tools/debugging-tools.ts#L89-L91). In `getDebuggingTools`, at the [`query_logs` registration](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/tools/debugging-tools.ts#L181-L197):

```ts
query_logs: injectableTool({
  ...debuggingToolDefs.query_logs,
  description:
    debugging.logsDialect === 'bigquery'
      ? debuggingToolDefs.get_logs.description
      : debuggingToolDefs.query_logs.description,
  inject: { project_id },
  execute: async (...) => { ... },
}),
```

The same whole-string switch applies to any other dialect-gated copy (e.g. the [`sql` param hint](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/tools/debugging-tools.ts#L40-L45), which today hardcodes the ClickHouse `logs` table / `log_attributes[...]` schema).

**4. The gating you already wrote does the rest.** [`hidden: Boolean(queryLogs)`](https://github.com/supabase/mcp/blob/9187f41195d40818868668d1476d653285f21e06/packages/mcp-server-supabase/src/tools/debugging-tools.ts#L161-L165) stays exactly as-is — once every platform implements `queryLogs`, `get_logs` is hidden everywhere and `query_logs` is the single listed logs tool on all envs, each with a dialect-appropriate description. No new environment conditionals beyond the one `logsDialect` field.

Not blocking on this, happy to ship #333 as-is to keep the ball rolling. Just flagging it now since it looks like a small, self-contained change that removes the hosted-vs-self-hosted split entirely. Wdyt?

PR #334, opened 2026-07-17
feat: hide tools from tools/list

gregnr:
Adds a `hidden: true` flag to tool definitions that will hide the tool from `tools/list` calls but allow it to still be callable via `tools/call`. Needed for the [expand+contract](https://github.com/supabase/mcp/blob/main/CONTRIBUTING.md#breaking-changes) breaking change pattern.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,10 @@ Follow the [expand and contract pattern](https://martinfowler.com/bliki/Parallel
 - **Expand:** add new tools/fields alongside the old ones. New fields must be optional, so clients still on the frozen schema keep validating. Don't touch the optionality of existing fields.
 - **Contract:** only remove the old tools/fields once the new version is published, so no frozen client can still depend on them.
 
+**Version the break at contract, not expand**
+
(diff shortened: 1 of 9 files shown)


mattrossman:
We might want to update `createToolSchemas()` / `supabaseMcpToolSchemas` to respect this hidden field and/or have a test to make the expectation for how it behaves clear.

Otherwise, [this test](https://github.com/supabase/mcp/blob/736277e2a9ccfcab871e3f356cf7a57dec9f3be1/packages/mcp-server-supabase/src/server.test.ts#L3115-L3153) might fail once we mark a tool as hidden since there will be tools in the schema registry that aren't in `tools/list`.

mattrossman (approved):

gregnr:
@mattrossman I updated the test to actually allow hidden tools to appear in `createToolSchemas()` (rather than filtering). The reason is that `hidden: true` is added during the expand phase, which is non-breaking from a semver POV (it just stops clients from discovering the tool at wire-level via `tools/list`). If we hide the tool from `createToolSchemas()`, that is breaking from a consumer's POV when our semver hasn't reflected that yet. Instead we make the breaking changes in the contract phase, which fully removes the tool.

I'll merge as is for now but happy to discuss further/change if you disagree with this.

PR #339, opened 2026-07-22
fix: stop double JSON-encoding `execute_sql` and `get_logs` results

### LINKED ISSUE #311: execute_sql corrupts backslash escaping in PL/pgSQL function bodies (round-trip failure)
# Bug report

## Describe the bug
When using the MCP `execute_sql` tool to `CREATE OR REPLACE FUNCTION` with PL/pgSQL bodies containing E-string backslash literals (e.g., `E'\\'`), the stored function body has incorrect escaping. The backslashes get doubled at some point in the JSON → MCP → PostgreSQL pipeline, resulting in a function that is semantically different from what was submitted.

This makes it impossible to reliably round-trip function definitions through `execute_sql`:

1. Read a function via `SELECT pg_get_functiondef(oid)` using `execute_sql`
2. Use the returned definition in a `CREATE OR REPLACE FUNCTION` via `execute_sql`
3. The resulting function body has corrupted backslash sequences

## To Reproduce
1. Have a function on your Supabase project containing an E-string backslash literal:
```sql
-- This function compares the first char against a literal backslash
IF left(NEW.access_token, 1) != E'\\' THEN
```
2. Read the function definition via MCP execute_sql:
```sql
SELECT pg_get_functiondef(oid) FROM pg_proc WHERE proname = 'encrypt_token';
```
3. Take the returned `CREATE OR REPLACE FUNCTION` statement and execute it via MCP execute_sql.
4. Compare the stored function body:
```sql
SELECT md5(pg_get_functiondef(oid)) FROM pg_proc WHERE proname = 'encrypt_token';
```
5. The hash does not match the original. Inspecting the body reveals the E-string `E'\\'` was stored as `E'\\\\'` (two backslashes instead of one), and comment text like `(\x...)` became `(\\x...)`.

## Expected behavior
The `CREATE OR REPLACE FUNCTION` statement executed via `execute_sql` should produce a byte-for-byte identical function body to the one that was read. Backslash sequences in dollar-quoted function bodies (`$function$...$function$`) should be passed through to PostgreSQL verbatim without additional escaping.

## Workaround
Using a different dollar-quote tag (e.g., `$fn_body$`) and manually adjusting the backslash count can produce the correct result, but it requires trial-and-error to determine how many levels of escaping the MCP layer adds. This is fragile and not suitable for automated schema replication workflows.

## Environment
MCP server: `https://mcp.supabase.com/mcp`
PostgreSQL version: 17.6.1.113
Agent: Claude Opus (via OpenCode)
Observed during branch provisioning (replicating function definitions from prod to a new branch via MCP `execute_sql`)

radmirnovii:
Fixes supabase/mcp#311

## Root cause

As diagnosed in [#311 (comment)](<https://github.com/supabase/mcp/issues/311#issuecomment-4848223834>), `execute_sql` results were JSON-encoded twice:

1. `wrapWithUntrustedDataBoundary` embeds the query rows via `JSON.stringify` inside the `<untrusted-data-*>` prose wrapper.
2. `createMcpServer`'s tool-call handler then serializes the whole output object (`{ result: "<wrapper text>" }`) with a second `JSON.stringify` to produce the text content.

A single backslash in queried data (e.g. the `E'\\'` literal in a PL/pgSQL function body) reached the model as 8 backslashes. Models routinely collapse that to the wrong count when writing the definition back, corrupting round-trips of function definitions. `get_logs` had the same double encoding.

## Fix

Tools in `@supabase/mcp-utils` can now declare an optional `textContent(output)` serializer:

* The returned text is sent **verbatim** as the text content (no second `JSON.stringify`).
* The full output object is also returned as [`structuredContent`](<https://modelcontextprotocol.io/specification/2025-11-25/server/tools#structured-content>), so typed clients keep parsing outputs exactly as before — the AI SDK integration (`supabaseMcpToolSchemas`) validates `structuredContent` first and never falls back to parsing the prose text.

`execute_sql` and `get_logs` use it to send the untrusted-data wrapper directly, so query data is now JSON-encoded exactly once. All other tools are unchanged.

This is also a first concrete step toward the `outputSchema`/`structuredContent` support mentioned in the issue: `createMcpServer` now emits `structuredContent` (currently only for tools that opt into `textContent`); advertising `outputSchema` in `tools/list` and emitting `structuredContent` universally can build on top of this.

## Behavior change

* `execute_sql` / `get_logs` text content is now the untrusted-data wrapper itself (prose containing single-encoded JSON) instead of a JSON object with the wrapper double-encoded inside. The text is meant for model consumption and was never parseable into row data in one step before.
* Both tools' results now include `structuredContent` (`{ result: string }`).
* `tools/list` output is untouched — no changes to tool names, descriptions, input schemas, or annotations (the frozen-field contract for ChatGPT snapshots is unaffected; `outputSchema` is not advertised in `tools/list`).
* `get_logs`'s internal output schema is corrected from `{ result: z.unknown() }` to `{ result: z.string() }` to match what it has always returned.

## Verification

* New regression test reproduces the exact scenario from the issue against PGlite: create a PL/pgSQL function containing `E'\\'`, read it back via `pg_get_functiondef` through `execute_sql`, assert the text contains the E-string encoded exactly once (4 backslashes, not 8), re-create the function from the returned definition, and assert the stored body is byte-identical. Also asserts `structuredContent` matches the text.
* New `mcp-utils` test covers `textContent` verbatim text + `structuredContent` emission.

```
pnpm --filter @supabase/mcp-utils test        # 13 passed (13)
pnpm --filter @supabase/mcp-server-supabase test:unit  # 192 passed (192)
```

Typecheck (`tsc --noEmit`) and `biome check` pass in both packages.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ const tools = await mcpClient.tools({
 ```
 
 > [!NOTE]
-> Most tools in this server do not send `structuredContent` in MCP tool results, so AI SDK falls back to parsing JSON from `content` text. The exceptions are `execute_sql` and `get_logs`, which do send `structuredContent` — AI SDK validates it directly. Their `content` text is prose (an untrusted-data wrapper around JSON) rather than a JSON object.
+> Most tools in this server do not send `structuredContent` in MCP tool results, so AI SDK falls back to parsing JSON from `content` text. The exceptions are `execute_sql`, `get_logs`, and `query_logs`, which do send `structuredContent` — AI SDK validates it directly. Their `content` text is prose (an untrusted-data wrapper around JSON) rather than a JSON object.
 
 For more information, see [Schema Definition](https://ai-sdk.dev/docs/ai-sdk-core/mcp-tools#schema-definition) and [Typed Tool Outputs](https://ai-sdk.dev/docs/ai-sdk-core/mcp-tools#typed-tool-outputs) in the AI SDK docs.
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ const tools = await mcpClient.tools({
 ```
 
 > [!NOTE]
-> This server does not send `structuredContent` in MCP tool results. AI SDK falls back to parsing JSON from `content` text.
+> Most tools in this server do not send `structuredContent` in MCP tool results, so AI SDK falls back to parsing JSON from `content` text. The exceptions are `execute_sql` and `get_logs`, which do send `structuredContent` — AI SDK validates it directly. Their `content` text is prose (an untrusted-data wrapper around JSON) rather than a JSON object.
 
 For more information, see [Schema Definition](https://ai-sdk.dev/docs/ai-sdk-core/mcp-tools#schema-definition) and [Typed Tool Outputs](https://ai-sdk.dev/docs/ai-sdk-core/mcp-tools#typed-tool-outputs) in the AI SDK docs.
 
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -2170,6 +2170,56 @@ describe('tools', () => {
     );
   });
 
+  test('get_logs encodes results exactly once and returns structured content', async () => {
+    const { client } = await setup();
+
+    const org = await createOrganization({
+      name: 'My Org',
+      plan: 'free',
+      allowed_release_channels: ['ga'],
+    });
+
+    const project = await createProject({
+      name: 'Project 1',
+      region: 'us-east-1',
+      organization_id: org.id,
+    });
+    project.status = 'ACTIVE_HEALTHY';
+
+    const logs = [{ event_message: String.raw`error reading C:\temp\file` }];
+
+    mockServer?.use(
+      http.get<{ projectId: string }>(
+        `${API_URL}/v1/projects/:projectId/analytics/endpoints/logs`,
+        () => HttpResponse.json(logs)
+      )
+    );
+
+    const output = await client.callTool({
+      name: 'get_logs',
+      arguments: {
+        project_id: project.id,
+        service: 'api',
+      },
+    });
+
+    const result = CallToolResultSchema.parse(output);
+    const [textContent] = result.content;
+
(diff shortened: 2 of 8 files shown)


barryroodt (commented):
Thanks for the contribution, and for digging into this. The double-encoding diagnosis is exactly right, and I appreciate that you traced it to the two `JSON.stringify` calls instead of patching backslash counts at the string level.

I checked it locally rather than just reading the diff: both suites pass at head (`mcp-utils` 13/13, supabase unit 192/192), and the merge-base encoding does fail your new test, so it's a real regression test. Re-creating the function from the returned definition and comparing the stored body was a nice touch.

One small request. [README.md:113-114](https://github.com/supabase/mcp/blob/2101930f50e29ce22e6c90f5d5c89eb48fcecbeb/README.md#L113-L114) says this server doesn't send `structuredContent`, and that AI SDK falls back to parsing JSON from `content` text. Both halves stop being true for `execute_sql` and `get_logs` with your change, and that note sits in the canonical AI SDK integration section, so an integrator reading it would get the opposite of what the server now does. If you're up for amending those two lines here, great. If you'd rather keep the PR to the fix, that's fine too and one of us can follow up with the docs.

I also have some non-blocking notes from a deeper read, an `injectableTool` cleanup plus a couple of coverage gaps. Happy to post them once this lands, or now if you'd prefer them together.

radmirnovii:
> One small request. [README.md:113-114](https://github.com/supabase/mcp/blob/2101930f50e29ce22e6c90f5d5c89eb48fcecbeb/README.md#L113-L114) says this server doesn't send `structuredContent`, and that AI SDK falls back to parsing JSON from `content` text. Both halves stop being true for `execute_sql` and `get_logs` with your change, and that note sits in the canonical AI SDK integration section, so an integrator reading it would get the opposite of what the server now does. If you're up for amending those two lines here, great. If you'd rather keep the PR to the fix, that's fine too and one of us can follow up with the docs.

fixed + added tests, sorry for long reply

PR #341, opened 2026-07-22
feat: add query_logs tool for custom log queries (CI mirror of #333)

jordienr:
Same changes as #333, pushed to a branch on the repo (not a fork) so the e2e tests run with CI secrets — #333's `test` check can't pass from the fork because forks don't receive secrets.

Do not merge this in place of #333 unless we decide to; it exists to verify the e2e suite goes green with secrets available. #333 remains the review PR.

Identical head commit as #333 (`c1bdc7f`).


barryroodt:
Structured review done against this head (c1bdc7f, same SHA as #333, so everything transfers). CI green here confirms the #333 failures were fork-secrets only. The `resolveLogWindow` fix is right, and `server.test.ts:1718` pins the end-anchored contract properly.

**One blocker before the checkmark: the types regen drops fields that exist on main.** `management-api/types.ts` at this head removes 10 `id: string` fields from the SAML provider/domain response shapes vs current `origin/main`. Stale OpenAPI snapshot by the look of it (the Pgsodium/PlanGate hunks are fine, already on main). Fix is mechanical: rebase, re-run the regen.

Optional while you're in there: `resolveLogWindow` throws a raw `Invalid time value` on a malformed `iso_timestamp_end`. Caught by the framework, but a parseability check (and `start < end`) would give models a clear validation error instead.

Rodriguespn:
Closing this one as https://github.com/supabase/mcp/pull/333 was merged

PR #343, opened 2026-07-23
feat: add --content-api-url flag and SUPABASE_CONTENT_API_URL env var

barryroodt:
Lets the stdio server point the docs Content API client at an alternative GraphQL endpoint instead of the hosted `https://supabase.com/docs/api/graphql` default. Precedence: flag, then env var, then the existing default. Behavior without either is unchanged (the default still applies, and no request is made anywhere new).

**Why**: we are wiring the evals harness so a docs change can be tested end to end before it ships: build a local docs content index, point `search_docs` at it, and run the affected evals against the local mcp build. That needs a supported way to override the Content API endpoint at launch. It also helps anyone running the server against a staging or self-hosted docs deployment.

**What's in here**

* `--content-api-url` CLI flag and `SUPABASE_CONTENT_API_URL` env var on the stdio transport, threaded to the existing `contentApiUrl` option of `createSupabaseMcpServer` (which already carries the hosted default).
* Integration tests (`test/stdio.integration.ts`): the spawned server routes content API traffic to the given URL (asserted by round-tripping the schema into the `search_docs` tool description from a local HTTP stub), the env var works as the fallback, and the flag wins when both are set.

Verified locally: `vitest run --project integration` 5/5, `tsc --noEmit` clean.

### DIFF
--- a/packages/mcp-server-supabase/src/transports/stdio.ts
+++ b/packages/mcp-server-supabase/src/transports/stdio.ts
@@ -16,6 +16,7 @@ async function main() {
       ['project-ref']: projectId,
       ['read-only']: readOnly,
       ['api-url']: apiUrl,
+      ['content-api-url']: cliContentApiUrl,
       ['version']: showVersion,
       ['features']: cliFeatures,
     },
@@ -32,6 +33,9 @@ async function main() {
         default: false,
       },
       ['api-url']: {
+        type: 'string',
+      },
+      ['content-api-url']: {
         type: 'string',
       },
       ['version']: {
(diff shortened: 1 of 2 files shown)


barryroodt:
CI status: the `quality` failure was a formatting slip in the new integration test, fixed and pushed. The remaining `check` failure is the daily management API types drift, not this PR: the sync workflow opened #340 with the regenerated types, so `check` goes green here once that merges. (`test` also flaked once on the pre-existing `generatePassword` symbols test, passed on rerun.)

Rodriguespn (approved):
Thanks for working on this ✅

PR #349, opened 2026-07-29
fix: hide read-only mode

### LINKED ISSUE #281: read-only stdio mode still exposes mutating tools in tools/list

raulb:

Fixes https://github.com/supabase/mcp/issues/281
Fixes https://linear.app/supabase/issue/AI-736/read-only-stdio-mode-still-exposes-mutating-tools-in-toolslist

### DIFF
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -3616,41 +3616,6 @@ describe('tools', () => {
     ).toEqual([]);
   });
 
-  test('all write tools (readOnlyHint: false) are excluded from readOnly schemas', async () => {
-    // Use readOnly server so dynamic annotations (e.g. execute_sql) reflect
-    // read-only mode correctly — execute_sql reports readOnlyHint: true when
-    // the server is read-only, so it won't be incorrectly caught by the filter below.
-    const { client } = await setup({
-      readOnly: true,
-      features: [
-        'docs',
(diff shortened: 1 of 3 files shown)


packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -3682,6 +3682,51 @@ describe('tools', () => {
     const parsedContent = JSON.parse(firstContent.text);
     expect(parsedContent).toBeTypeOf('object');
   });
+
```
barryroodt, line 3650: [server.test.ts:3619](https://github.com/supabase/mcp/blob/03fa98760a91cb6ee6ac9f1736983ad7b364af09/packages/mcp-server-supabase/src/server.test.ts#L3619) can no longer fail. 

It loops over write tools found in `listTools()`, which is now always empty (your new test asserts that), so the body never runs. I'd just delete it, the new test covers it.

barryroodt (approved):
LGTM - just a small non-blocking note/nit

PR #357, opened 2026-08-10
fix: select query_logs dialect via logsDialect

Rodriguespn:
Adds an optional `logsDialect` (`'clickhouse' | 'bigquery'`) to `DebuggingOperations` so each platform declares which SQL its logs endpoint speaks. `query_logs` then picks its whole dialect-appropriate description and `sql` param hint from a single lookup table, instead of hardcoding ClickHouse copy — no environment-conditional prose the model has to reason about. Defaults to `'clickhouse'` when unset, so existing platform implementers are unaffected.

Implements [Matt's `logsDialect` suggestion](https://supabase.slack.com/archives/C08N7894QTG/p1784660787397029?thread_ts=1784624411.491679&cid=C08N7894QTG) raised in the #333 review.

**Why**: it lets `query_logs` roll out across every environment without env-specific branching. Hosted stays ClickHouse; self-hosted/CLI can keep its BigQuery-backed logs endpoint and declare `logsDialect: 'bigquery'`. This unblocks shipping the MCP server to self-hosted (studio) regardless of whether it's on BigQuery or ClickHouse, rather than holding the studio image back until self-hosted moves to ClickHouse.

> [!NOTE]
> The comment sketched "reuse the `get_logs` description for `bigquery`" — I instead give `query_logs` a proper BigQuery-dialect description, since `get_logs` describes a `service` param while `query_logs` takes `sql`. The BigQuery `sql` hint mirrors the canonical Logflare self-hosted schema (`cross join unnest(metadata)`); worth confirming when studio wires up the platform.

Refs AI-1046

### DIFF
--- a/packages/mcp-server-supabase/src/tools/debugging-tools.ts
+++ b/packages/mcp-server-supabase/src/tools/debugging-tools.ts
@@ -58,27 +58,48 @@ function buildQueryLogsInputSchema(sqlDescription: string) {
 }
 
 /**
- * The `query_logs` tool copy, keyed by the dialect its logs endpoint speaks.
- * The whole description and `sql` param hint are swapped per dialect — no
- * conditional prose the model has to reason about. Precomputed at module level
- * because zod schemas with `.describe()` auto-register in the global registry,
- * so rebuilding them per `getDebuggingTools` call would grow it unboundedly.
+ * Builds the `query_logs` copy for one dialect from the shared wording, so only
+ * the two parts that genuinely differ per dialect are supplied: the SQL dialect
+ * `name` and the `schemaHint` describing how that backend exposes logs.
+ */
+function buildQueryLogsCopy({
+  name,
+  schemaHint,
+}: {
+  name: string;
+  schemaHint: string;
+}) {
+  return {
+    description: `Runs a custom read-only ${name} SQL query against a Supabase project's unified logs stream, for filtering, aggregating, or joining across log fields more precisely than a simple per-service log dump. When the user asks about a specific time range, always pass iso_timestamp_start and iso_timestamp_end to match it; otherwise the query defaults to the last 24 hours and will return results from a wider window than intended. The window can be up to 24 hours. Do not poll this tool in a loop.`,
+    parameters: buildQueryLogsInputSchema(
+      `A read-only ${name} SQL query to run against the project's unified logs stream. ${schemaHint}`
+    ),
+  };
+}
+
+/**
+ * The `query_logs` tool copy, keyed by the dialect its logs endpoint speaks. The
+ * whole description and `sql` param hint are swapped per dialect — no conditional
+ * prose the model has to reason about. `schemaHint` mirrors how each backend
+ * exposes logs: hosted ClickHouse serves a single `logs` table filtered by
+ * `source`, while self-hosted BigQuery (Logflare) serves one source table per
+ * service with fields nested under `metadata` (see
+ * `apps/studio/lib/api/self-hosted/logs.ts` in `supabase`). Precomputed at module
+ * level because zod schemas with `.describe()` auto-register in the global
+ * registry, so rebuilding them per `getDebuggingTools` call would grow it
+ * unboundedly.
  */
 const queryLogsByDialect = {
-  clickhouse: {
-    description:
-      "Runs a custom read-only ClickHouse SQL query against a Supabase project's unified logs stream, for filtering, aggregating, or joining across log fields more precisely than a simple per-service log dump. When the user asks about a specific time range, always pass iso_timestamp_start and iso_timestamp_end to match it; otherwise the query defaults to the last 24 hours and will return results from a wider window than intended. The window can be up to 24 hours. Do not poll this tool in a loop.",
-    parameters: buildQueryLogsInputSchema(
-      "A read-only ClickHouse SQL query to run against the project's unified logs stream. Logs are exposed through a `logs` table; filter by `source` (e.g. 'edge_logs', 'postgres_logs', 'function_edge_logs', 'function_logs', 'auth_logs', 'storage_logs', 'realtime_logs', 'workflow_run_logs') and read nested fields via `log_attributes['<key>']`."
-    ),
-  },
-  bigquery: {
-    description:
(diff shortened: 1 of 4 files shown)


packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -35,27 +38,75 @@ const getLogsOutputSchema = z.object({
...
+      ),
+  });
+}
+
+/**
+ * Builds the `query_logs` copy for one dialect from the shared wording, so only
+ * the two parts that genuinely differ per dialect are supplied: the SQL dialect
+ * `name` and the `schemaHint` describing how that backend exposes logs.
+ */
+function buildQueryLogsCopy({
```
mattrossman, line 66: The `debuggingToolDefs` -> `buildQueryLogsCopy` -> `buildQueryLogsInputSchema` chain is a little awkward for me to follow, it'd be easier for me to read flat factories that convert `dialect -> description` and `dialect -> parameters` called wherever description/parameters are needed but not a dealbreaker if you prefer as-is (esp since logsDialect is only needed short term)

packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -35,27 +38,75 @@ const getLogsOutputSchema = z.object({
...
+ * `apps/studio/lib/api/self-hosted/logs.ts` in `supabase`). Precomputed at module
+ * level because zod schemas with `.describe()` auto-register in the global
+ * registry, so rebuilding them per `getDebuggingTools` call would grow it
+ * unboundedly.
+ */
+const queryLogsByDialect = {
+  clickhouse: buildQueryLogsCopy({
+    name: 'ClickHouse',
+    schemaHint:
+      "Logs are exposed through a `logs` table; filter by `source` (e.g. 'edge_logs', 'postgres_logs', 'function_edge_logs', 'function_logs', 'auth_logs', 'storage_logs', 'realtime_logs', 'workflow_run_logs') and read nested fields via `log_attributes['<key>']`.",
```
mattrossman: Unrelated, but in my testing I noticed a `postgrest_logs` table available in staging not part of this list.

I gather this is just an example not meant to be exhaustive, but I wonder:
- Will agents reliably discover the list of available sources before querying? Is there risk of hallucination if our description doesn't match the real full schema?
- Will they understand what each log source represents? For example, will they know what `function_edge_logs` means compared to `function_logs`?

Not something we need to solve in this PR but maybe worth revisiting w/ evals.

mattrossman (approved):
Tested the preview build working against staging. Agreed this `logsDialect` approach adds flexibility for easier migration to `query_logs` across environments.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
Rodriguespn: I'll leave it as it is as this is a short-term solution. Thank you for the suggestion regardless.

packages/mcp-server-supabase/src/tools/debugging-tools.ts
Rodriguespn: Thanks @mattrossman. I dug in and made a change plus stood up evals to keep us honest about it. Result below.

The log set is genuinely incomplete. I checked a live self-hosted Logflare catalog (`_analytics.sources`) and it exposes `postgREST.logs.prod` **and** `pgbouncer.logs.prod`, neither of which we listed. So I trimmed it to a few representative examples marked non-exhaustive, and added a discovery step, for both dialects:

**ClickHouse — from:**
> A read-only ClickHouse SQL query to run against the project's unified logs stream. Logs are exposed through a `logs` table; filter by `source` (e.g. 'edge_logs', 'postgres_logs', 'function_edge_logs', 'function_logs', 'auth_logs', 'storage_logs', 'realtime_logs', 'workflow_run_logs') and read nested fields via `log_attributes['<key>']`.

**to:**
> A read-only ClickHouse SQL query to run against the project's unified logs stream. Logs are exposed through a `logs` table; filter by `source` (common values include 'edge_logs', 'postgres_logs', and 'function_edge_logs', but this list is not exhaustive — run `select distinct source from logs` to discover the sources available for this project) and read nested fields via `log_attributes['<key>']`.

**BigQuery — from:**
> A read-only BigQuery SQL query to run against the project's unified logs stream. Each service has its own source table (e.g. 'edge_logs', 'postgres_logs', 'function_edge_logs', 'auth_logs', 'storage_logs', 'realtime_logs'); nested fields live under `metadata`, read via `cross join unnest(metadata) as m` and then `m.<field>` (unnest further for nested structs such as `m.request` or `m.parsed`).

**to:**
> A read-only BigQuery SQL query to run against the project's logs. Each service has its own source table (common ones include 'edge_logs', 'postgres_logs', and 'function_edge_logs', but the set is not exhaustive — other per-service tables, e.g. Data API and pooler logs, also exist); read nested fields by cross joining `unnest(metadata) as m` and then `m.<field>` (unnest further for nested structs such as `m.request` or `m.parsed`).

(BigQuery has no unified stream to `select distinct source` from, so it names the real-but-unlisted families instead. I also fixed "unified logs stream" → "logs" there, since it's per-table.)


---


### Created 2 evals to answer both questions

Two "ephemeral" regression evals, pinned to this PR's preview build: **supabase/evals#185**. [CI run](https://github.com/supabase/evals/actions/runs/31417243638) across gpt-5.4-mini + claude-code sonnet 5 — **all 4 pairs pass**:

| Eval | gpt-5.4-mini (codex) | claude-code sonnet 5 |
|---|---|---|
| logs-002 (undescribed source) | ✅ 3/3 | ✅ 3/3 |
| logs-003 (function streams) | ✅ 2/3 | ✅ 3/3 |

Rodriguespn: Answering directly to your questions:

> do agents discover sources / is there hallucination risk?

`investigate-logs-002-undescribed-source-discovery`: the root cause (`permission denied for table orders`, 403) lives **only in `postgrest_logs`**, which the description never lists. 

Result: both models solved it and **neither hallucinated**
- both first *enumerated* the available sources (`select … group by source` / `distinct source`)
- then queried `postgrest_logs` **by name**.  

So they don't blindly trust the listed set; given the unified stream they discover what's actually there.


> do they understand `function_edge_logs` vs `function_logs`?

`investigate-logs-003-function-stream-disambiguation`: the bug (charging dollars as cents) is only in the runtime console (`function_logs`), while the request stream (`function_edge_logs`) shows healthy 200s.

Result: both models surfaced the runtime bug and **weren't fooled** by the 200s. 

- claude-code sonnet 5 specifically queried the `function_logs` stream (understood the distinction, 3/3); 
- gpt-5.4-mini reached it by content-filtering the unified stream rather than naming the source (2/3).

PR #358, opened 2026-08-11
feat: add dual-era package serving entries

barryroodt:
> **This is a stacked PR. Its parent is #327, please review that one first.**
>
> The diff here is only this branch's own commits. The base is `raulb/spike-mcp-v2` rather than `main` on purpose, because #327 is still in review; once it lands, this retargets to `main` and the diff against `main` shrinks accordingly.
>
> The `test` check runs on this page and passes at `851f01e`. An earlier version of this description said it would be absent; that note is out of date. The test output under Verification came from a local run that predates the check.
>
> At merge time: retarget to `main`, let `tests.yml` run green there, then merge.

The CLI serves both protocol eras from one tool definition, and the package exports one modern HTTP handler for the hosted `/mcp` endpoint to mount.

PR 2 of 3 for [AI-1044](https://linear.app/supabase/issue/AI-1044/migrate-mcp-server-supabase-to-mcp-sdk-v2) ([plan](https://linear.app/supabase/document/mcp-sdk-v2-migration-plan-d19bb064d673)).

`src/transports/stdio.ts` connects a raw `StdioServerTransport`, so only 2025-era clients are served, and there's no HTTP entry for hosted to consume.

## What is the new behavior?

`serveStdio` owns the transport and picks the era per connection. `createSupabaseMcpHandler` wraps the SDK's `createMcpHandler` with `legacy: 'reject'`, so hosted keeps authentication, ABAC, era dispatch, and logging.

`tools/list` is unchanged, so the frozen ChatGPT contract is unaffected. `mcp-server-postgrest` keeps its single-era entry, and nothing here adds Elicitation machinery, tool policies, or telemetry.

## Verification

- `mcp-server-supabase` unit + integration: **222 passed**. Both eras run against the real built `dist/transports/stdio.js`, spawned as a child process.
- `mcp-utils`: **12 passed**. `pnpm build` and `pnpm format:check` clean.
- `pnpm test:packed-platform-consumer` (new): **3/3** on packed `0.10.0`. Installs the tarballs with plain `npm` outside the repo on Platform's exact zod pin, then checks the CJS entry, a `tsc` typecheck of the packed declarations, and one modern call that imports the ESM entry.
- CLI behavior unchanged: `--version` exits 0, a missing access token exits 1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
     "build": "pnpm --filter @supabase/mcp-utils --filter @supabase/mcp-server-supabase --filter @supabase/mcp-server-postgrest build",
     "test": "pnpm --parallel --filter @supabase/mcp-utils --filter @supabase/mcp-server-supabase --filter @supabase/mcp-server-postgrest test",
     "test:coverage": "pnpm --filter @supabase/mcp-server-supabase test:coverage",
+    "test:packed-platform-consumer": "node scripts/test-packed-platform-consumer.mjs",
     "format": "biome check --write .",
     "format:check": "biome check ."
   },
--- a/packages/mcp-server-supabase/src/transports/http.test.ts
+++ b/packages/mcp-server-supabase/src/transports/http.test.ts
@@ -1,0 +1,231 @@
+import {
+  Client,
+  StreamableHTTPClientTransport,
+} from '@modelcontextprotocol/client';
+import {
+  CLIENT_CAPABILITIES_META_KEY,
+  PROTOCOL_VERSION_META_KEY,
+} from '@modelcontextprotocol/server';
+import { http, HttpResponse } from 'msw';
+import { setupServer, type SetupServer } from 'msw/node';
+import { afterEach, beforeEach, describe, expect, test } from 'vitest';
+
+import {
+  ACCESS_TOKEN,
+  API_URL,
+  MCP_CLIENT_NAME,
+  MCP_CLIENT_VERSION,
+  mockBranches,
+  mockContentApi,
+  mockContentApiSchemaLoadCount,
+  mockManagementApi,
+  mockOrgs,
+  mockProjects,
+} from '../../test/mocks.js';
+import { createSupabaseApiPlatform } from '../platform/api-platform.js';
+import { createSupabaseMcpHandler } from './http.js';
+
+const MODERN_PROTOCOL_VERSION = '2026-07-28';
+const MCP_ENDPOINT = new URL('https://mcp.test');
+
+let mockServer: SetupServer | undefined;
+const cleanups: Array<() => Promise<void>> = [];
+
+beforeEach(() => {
+  mockOrgs.clear();
+  mockProjects.clear();
+  mockBranches.clear();
+  mockContentApiSchemaLoadCount.value = 0;
+
+  mockServer = setupServer(...mockContentApi, ...mockManagementApi);
+  mockServer.listen({ onUnhandledRequest: 'error' });
+});
+
+afterEach(async () => {
+  try {
+    for (const cleanup of cleanups.splice(0).reverse()) {
+      await cleanup();
+    }
+  } finally {
+    mockServer?.close();
+  }
+});
+
+function createPlatform() {
+  return createSupabaseApiPlatform({
+    accessToken: ACCESS_TOKEN,
+    apiUrl: API_URL,
+  });
+}
+
+function createHandler(platform = createPlatform()) {
+  const handler = createSupabaseMcpHandler({ platform, readOnly: true });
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
     "build": "pnpm --filter @supabase/mcp-utils --filter @supabase/mcp-server-supabase --filter @supabase/mcp-server-postgrest build",
     "test": "pnpm --parallel --filter @supabase/mcp-utils --filter @supabase/mcp-server-supabase --filter @supabase/mcp-server-postgrest test",
     "test:coverage": "pnpm --filter @supabase/mcp-server-supabase test:coverage",
+    "test:packed-platform-consumer": "node scripts/test-packed-platform-consumer.mjs",
```
raulb, line 6: I wonder if in some point we could generalize this to run it in a workflow to ensure compatibility with platform.

packages/mcp-server-supabase/src/transports/http.ts
```diff
@@ -0,0 +1,12 @@
+import { createMcpHandler } from '@modelcontextprotocol/server';
+
+import {
+  createSupabaseMcpServer,
+  type SupabaseMcpServerOptions,
+} from '../server.js';
+
+export function createSupabaseMcpHandler(options: SupabaseMcpServerOptions) {
+  return createMcpHandler(() => createSupabaseMcpServer(options), {
+    legacy: 'reject',
```
raulb, line 12: I would add a comment explaining what we have in the readme.

Something along the lines of `Modern protocol only...`

packages/mcp-server-supabase/src/transports/http.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  ACCESS_TOKEN,
+  API_URL,
+  MCP_CLIENT_NAME,
+  MCP_CLIENT_VERSION,
+  setupMockApis,
+} from '../../test/mocks.js';
+import { createSupabaseApiPlatform } from '../platform/api-platform.js';
+import { createSupabaseMcpHandler } from './http.js';
+
+const MODERN_PROTOCOL_VERSION = '2026-07-28';
```
raulb, line 24: nit: add a link to https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/

raulb (approved):
@barryroodt left some comments, but my main question is the following one is re: the new scripts directory added. 

I believe that's because it's meant to be shared across `@supabase/mcp-server-supabase` and `@supabase/mcp-utils`. Is that right?

If so, I think it'd be important to state that in the CONTRIBUTING.md file since we're introducing a new pattern for these.

package.json
barryroodt: Agreed, and it's tracked: our own review of this branch flagged that the gate protects nothing until CI runs it. Left out of this PR to keep the diff to the migration itself, though that's a scope call rather than a hard rule.

Two halves:

**Wiring.** The whole script is 13s locally, and 14s with an empty npm cache. That includes its own `pnpm build` of both packages, which it runs unconditionally (`scripts/test-packed-platform-consumer.mjs:166`), so in `tests.yml` it would repeat a build the workflow already did. Cheap enough to leave alone, or worth a `--skip-build` flag. I'd put it there rather than a release step, so breaks land on the PR that causes them.

**The zod pin.** `PLATFORM_ZOD_VERSION` is a hardcoded `4.4.3` with an "update when Platform's catalog changes" comment, so it can drift from `develop` and stay green. `sync-management-api-types.yml` is already a daily cron of this shape.

Both halves are tracked in [AI-1075](https://linear.app/supabase/issue/AI-1075/wire-the-packed-consumer-gate-into-ci-and-detect-zod-pin-drift).

barryroodt (commented):
Yeah, that's right, it packs both packages and installs them together, so neither owns it. Root `package.json` already holds the cross-package commands (`build`, `test`). The check works because the consumer is built under `tmpdir()`, where `workspace:`, `catalog:` and symlinked `node_modules` can't reach it.

The new top-level `scripts/` dir does arrive unexplained though. I'll document it in CONTRIBUTING.md, slightly wider than you asked: there's no testing section at all today, so `pnpm test`, `test:coverage` and this gate are all undiscoverable.

packages/mcp-server-supabase/src/transports/http.test.ts
barryroodt: Thanks, will add.

packages/mcp-server-supabase/src/transports/http.ts
barryroodt: Good call, adding.

PR #390, opened 2026-08-28
feat: group lints in `get_advisors` response

mattrossman:
## Problem

`get_advisors` returns a list with one object per finding, each carrying the same keys.  The [lint SQL](https://github.com/supabase/supabase/blob/d5ee11bea0cba0149d16715aabb088d03c10be36/packages/pg-meta/src/sql/studio/advisor/lints.ts) emits `title`, `description` and `remediation` as literals on each row, so 300 instances of the same lint means those same three strings 300 times.

See diagnosis [thread](https://supabase.slack.com/archives/C051L8U2EJF/p1787774307978489) with [metrics](https://supabase.slack.com/archives/C051L8U2EJF/p1787777939555179?thread_ts=1787774307.978489&cid=C051L8U2EJF) demonstrating that is is the longest tail context hog in studio's AI Assistant.

## Fix

Groups findings by lint.

- Findings only merge when all seven [shared fields](https://github.com/supabase/mcp/pull/390/changes#diff-38cc0dc8b54f71f78652c17cbd5d29002718be0c4f3b0493e36f448f428e2b3eR132-R140) match
- Merged findings appear as a `findings` list on the `lints` entry, with convenience `count`
- Drops `cache_key`, which primarily exists for [the dashboard UI](https://github.com/supabase/supabase/blob/d5ee11bea0cba0149d16715aabb088d03c10be36/apps/studio/pages/project/%5Bref%5D/advisors/security.tsx#L64)

**Before**

```jsonc
{ "lints": [
  { "name": "lint_a", "title": "…", "level": "WARN", "facing": "…", "categories": ["…"], "description": "…", "remediation": "…",
    "detail": "… object_1 …", "metadata": { "name": "object_1" }, "cache_key": "…" },
  { "name": "lint_a", "title": "…", "level": "WARN", "facing": "…", "categories": ["…"], "description": "…", "remediation": "…",
    "detail": "… object_2 …", "metadata": { "name": "object_2" }, "cache_key": "…" },
  { "name": "lint_a", "title": "…", "level": "WARN", "facing": "…", "categories": ["…"], "description": "…", "remediation": "…",
    "detail": "… object_3 …", "metadata": { "name": "object_3" }, "cache_key": "…" },
  { "name": "lint_a", "title": "…", "level": "WARN", "facing": "…", "categories": ["…"], "description": "…", "remediation": "…",
    "detail": "… object_4 …", "metadata": { "name": "object_4" }, "cache_key": "…" },
  { "name": "lint_a", "title": "…", "level": "WARN", "facing": "…", "categories": ["…"], "description": "…", "remediation": "…",
    "detail": "… object_5 …", "metadata": { "name": "object_5" }, "cache_key": "…" },
  // ...296 more, each repeating the same 7 shared fields
]}
```

**After**

```jsonc
{ "lints": [
  { "name": "lint_a", "title": "…", "level": "WARN", "facing": "…", "categories": ["…"], "description": "…", "remediation": "…",
    "count": 301,
    "findings": [
      { "detail": "… object_1 …", "metadata": { "name": "object_1" } },
      { "detail": "… object_2 …", "metadata": { "name": "object_2" } },
      { "detail": "… object_3 …", "metadata": { "name": "object_3" } },
      { "detail": "… object_4 …", "metadata": { "name": "object_4" } },
      { "detail": "… object_5 …", "metadata": { "name": "object_5" } },
      // ...296 more
    ] }
]}
```

In the turn cited above, the response shrinks to 293,485 chars, ~73k tokens.

## How to review

Take a look at the new [tests](https://github.com/supabase/mcp/pull/390/changes#diff-4111b138e3c190a9303040787d584d08b2aa6b92a097b36bd47fa299c3008442) to understand the expecting grouping behavior.

Point an MCP client at the [preview](https://github.com/supabase/mcp/pull/390#issuecomment-5455072142) build below then call `get_advisors` for a project with some advisors, and observe the shape.

Screenshot from my testing in MCP Inspector:

CleanShot 2026-08-28 at 12 42 40@2x

## Follow-ups

- Even with the shrunken response, tool outputs are unbounded and can flood context. The [worst recent call](https://www.braintrust.dev/app/supabase.io/p/Assistant/trace?object_type=project_logs&object_id=5a8d02e5-b3b6-40cc-ba76-ecee286478f4&r=e5531007-e026-45b5-8dad-5426260bb2dd&s=4e233645-2b9e-4abf-820a-c5fd785945c9) returned 10,649 findings, where even with grouping it's ~2.5x over Assistant's context limit. This exploration will continue in [AI-865](https://linear.app/supabase/issue/AI-865/spike-large-tool-output-handling-in-assistant).
- `detail` is only included per-finding because `pg-meta` puts index names, policy names, roles and actions in prose instead of `metadata`. If we fill those in, `detail` could be dropped entirely, making that same payload ~4x smaller

`tools/list` output is untouched so no ChatGPT resubmission needed.

Closes AI-1136

### DIFF
--- a/packages/mcp-server-supabase/src/tools/debugging-tools.test.ts
+++ b/packages/mcp-server-supabase/src/tools/debugging-tools.test.ts
@@ -1,5 +1,18 @@
 import { describe, expect, test } from 'vitest';
-import { DAY_MS, resolveLogWindow } from './debugging-tools.js';
+import type { components } from '../management-api/types.js';
+import {
+  DAY_MS,
+  groupAdvisorLints,
+  groupAdvisorsResult,
+  resolveLogWindow,
+} from './debugging-tools.js';
+
+/**
+ * The lint shape the Management API actually returns, so these fixtures fail
+ * typecheck if `GET /v1/projects/{ref}/advisors/{type}` changes under us.
+ */
+type AdvisorLint =
+  components['schemas']['V1ProjectAdvisorsResponse']['lints'][number];
 
 describe('resolveLogWindow', () => {
   test('defaults the end to now and the start to 24 hours before it', () => {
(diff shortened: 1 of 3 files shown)


packages/mcp-server-supabase/src/tools/debugging-tools.ts
```diff
@@ -128,6 +128,95 @@ const getAdvisorsOutputSchema = z.object({
...
+      }
+      if (SHARED_LINT_FIELDS.includes(key)) {
+        shared[key] = value;
+      } else {
+        finding[key] = value;
+      }
+    }
+
+    const key = JSON.stringify(
+      SHARED_LINT_FIELDS.map((field) => shared[field] ?? null)
```
barryroodt: `shared[field] ?? null` gives a missing field, `undefined`, and explicit `null` the same key, so rows can merge without all seven fields matching. I reproduced this with two unrelated object-shaped lints, which became one group with `count: 2`. Could we validate the full lint shape before grouping, or encode field presence in the key?

packages/mcp-server-supabase/src/tools/debugging-tools.test.ts
```diff
@@ -69,3 +78,155 @@ describe('resolveLogWindow', () => {
...
+        'Detects functions where the search_path parameter is not set.',
+      detail: 'Function `public.a` has a role mutable search_path',
+      remediation: 'https://supabase.com/docs/guides/database/database-linter',
+      metadata: { schema: 'public', name: 'a', type: 'function' },
+      cache_key: 'function_search_path_mutable_public_a',
+      ...overrides,
+    };
+  }
+
+  test('collapses repeated lints into one entry', () => {
```
barryroodt, line 100: These tests call the exported helpers directly, while the existing `callTool` advisor tests only use empty lint arrays. Removing the production call at `debugging-tools.ts:362` would leave them green. Could we add a non-empty `get_advisors` tool-level test and keep the transformer private?

barryroodt (commented):
The payload reduction looks like a clear win. I’ve left two comments before approval: only group responses with the shape we expect, and cover grouping through the real `get_advisors` path.

barryroodt (approved):

packages/mcp-server-supabase/src/tools/debugging-tools.ts
mattrossman: Tested and addressed in https://github.com/supabase/mcp/pull/390/commits/4f18769d891bdd8e78fe489d66e3af1b463d0013

packages/mcp-server-supabase/src/tools/debugging-tools.test.ts
mattrossman: I added a more complete `get_advisors` tool test that includes perf & security advisors in https://github.com/supabase/mcp/pull/390/commits/ac196779658b947d90c5c850ff51adbbe5111133

I considered refactoring the other grouping unit tests too, but it adds a fair bit of boilerplate, so I'm gonna leave the more extensive unit tests here as tests against the direct helper (which we have least have precedent for from the `resolveLogWindow` helper/tests), with the broader tool-level test as the sanity check that the helper is being called.

PR #391, opened 2026-08-28
feat(mcp): add project cost confirmation elicitation

barryroodt:

Feature.

Implements the `create_project` slice of [AI-1091](https://linear.app/supabase/issue/AI-1091/ship-form-mode-cost-confirmation-stage-a-supabasemcp).

`create_project` requires a valid `confirm_cost_id` before execution. Clients without elicitation capability keep this flow.

## What is the new behavior?

When `costConfirmation` is configured with `enabledTools: ['create_project']`, form-capable clients can confirm the quoted cost inside `create_project` through the SDK-native `inputRequired` and `inputResponse` flow. The signed request state binds the method and authenticated principal, and carries `tool: 'create_project'`, the project name, region, organization, and quoted cost.

The form has no properties. Consent lives in the elicitation action, so clients use their native accept, decline, and cancel controls. Accept creates the project. Decline or cancel returns a normal result without creating anything.

Argument mismatches are rejected, and cost drift starts confirmation again. A state signed for the wrong tool fails before argument or acceptance checks. Empty elicitation mode maps count as form-capable; URL-only or absent elicitation uses the legacy `confirm_cost` flow. A precomputed legacy ID cannot bypass confirmation for a form-capable request.

This PR covers `create_project` only. It does not change stdio, and there is no `create_branch` implementation in this PR.

## How to Review

1. **Server setup**
   - `packages/mcp-server-supabase/src/server.ts`
   - Inspect how `costConfirmation`, its `enabledTools`, and request state enter the server.

2. **Confirmation flow and executor seam**
   - `packages/mcp-server-supabase/src/tools/cost-confirmation.ts`, `packages/mcp-server-supabase/src/tools/account-tools.ts`, `packages/mcp-server-supabase/src/tools/util.ts`, and `packages/mcp-utils/src/server.ts`
   - Inspect the private module that owns the project state discriminator, action-only schema, and capability helper, plus signed state, project creation behavior, request-state verification, and SDK result passthrough.

3. **Behavioral coverage**
   - `packages/mcp-server-supabase/src/server.test.ts`
   - Inspect the modern and legacy paths.

- [ ] Does capability detection select the right lane?
- [ ] Is signed state bound to the right request, tool, and principal?
- [ ] Do decline and cancel avoid project creation?
- [ ] Does the legacy flow remain strict?

## Verification

- `pnpm --filter @supabase/mcp-server-supabase test src/server.test.ts -t "create_project cost confirmation via elicitation"` passed 8 tests.
- `pnpm --filter @supabase/mcp-server-supabase build` passed.
- `pnpm test:packed-platform-consumer` passed all 3 public-surface assertions.
- `biome ci .` passed.

## Additional context

This is a fresh, project-only implementation separate from the earlier draft in [#389](https://github.com/supabase/mcp/pull/389).

When confirmation is configured, the public schema makes `confirm_cost_id` optional so form-capable clients can call `create_project` directly. Capability-free execution still requires a valid ID.

Known limits remain: signed state has no single-use replay prevention, and the existing final cost-check concurrency race remains.

The platform PR will first pin the pkg.pr.new preview, then switch to the released npm version.

This implementation was AI-assisted. The verification evidence is above, and human review is required before merge.

### DIFF
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -1,3 +1,4 @@
+import { createRequestStateCodec } from '@modelcontextprotocol/server';
 import {
   createMcpServer,
   type Tool,
@@ -6,7 +7,10 @@ import {
 import packageJson from '../package.json' with { type: 'json' };
 import { createContentApiClient } from './content-api/index.js';
 import type { SupabasePlatform } from './platform/types.js';
-import { getAccountTools } from './tools/account-tools.js';
+import {
+  getAccountTools,
+  type ProjectCostState,
+} from './tools/account-tools.js';
 import { getBranchingTools } from './tools/branching-tools.js';
 import { getDatabaseTools } from './tools/database-operation-tools.js';
 import { getDebuggingTools } from './tools/debugging-tools.js';
@@ -54,6 +58,21 @@ export type SupabaseMcpServerOptions = {
    * Callback for after a supabase tool is called.
    */
   onToolCall?: ToolCallCallback;
+
+  /**
+   * Enables cost confirmation via elicitation inside `create_project` for
+   * clients that declare per-request form-elicitation capability. Clients
+   * without that capability keep using `get_cost` -> `confirm_cost` ->
+   * `create_project(confirm_cost_id)`.
+   */
+  projectCostConfirmation?: {
+    /** HMAC key for the `requestState` codec. MUST be at least 32 bytes. */
+    requestStateKey: string | Uint8Array;
+    /** The authenticated principal `requestState` is bound to. */
+    principal: string;
+    /** How long a minted `requestState` stays valid, in seconds. */
+    ttlSeconds?: number;
+  };
 };
 
 const DEFAULT_FEATURES: FeatureGroup[] = [
@@ -96,6 +115,7 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
     features,
     contentApiUrl = 'https://supabase.com/docs/api/graphql',
     onToolCall,
+    projectCostConfirmation,
   } = options;
 
   const contentApiClientPromise = createContentApiClient(contentApiUrl, {
@@ -114,6 +134,15 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
     platform,
     features ?? availableDefaultFeatures
   );
+
+  const projectCostConfirmationCodec = projectCostConfirmation
+    ? createRequestStateCodec<ProjectCostState>({
+        key: projectCostConfirmation.requestStateKey,
+        ttlSeconds: projectCostConfirmation.ttlSeconds,
+        bind: (ctx) =>
+          `${ctx.mcpReq.method}:${projectCostConfirmation.principal}`,
+      })
+    : undefined;
 
   const server = createMcpServer({
     name: 'supabase',
@@ -134,6 +163,9 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
       ]);
     },
     onToolCall,
+    requestState: projectCostConfirmationCodec && {
+      verify: projectCostConfirmationCodec.verify,
+    },
     tools: async () => {
       const contentApiClient = await contentApiClientPromise;
       const tools: Record<string, Tool> = {};
@@ -153,7 +185,16 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
       }
 
       if (!projectId && account && enabledFeatures.has('account')) {
-        Object.assign(tools, getAccountTools({ account, readOnly }));
+        Object.assign(
+          tools,
+          getAccountTools({
+            account,
+            readOnly,
+            projectCostConfirmation: projectCostConfirmationCodec && {
+              codec: projectCostConfirmationCodec,
+            },
+          })
+        );
       }
 
       if (database && enabledFeatures.has('database')) {
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -7,11 +7,9 @@ import {
 import packageJson from '../package.json' with { type: 'json' };
 import { createContentApiClient } from './content-api/index.js';
 import type { SupabasePlatform } from './platform/types.js';
-import {
-  getAccountTools,
-  type ProjectCostState,
-} from './tools/account-tools.js';
+import { getAccountTools } from './tools/account-tools.js';
 import { getBranchingTools } from './tools/branching-tools.js';
+import type { CostConfirmationState } from './tools/cost-confirmation.js';
 import { getDatabaseTools } from './tools/database-operation-tools.js';
 import { getDebuggingTools } from './tools/debugging-tools.js';
 import { getDevelopmentTools } from './tools/development-tools.js';
@@ -60,18 +58,20 @@ export type SupabaseMcpServerOptions = {
   onToolCall?: ToolCallCallback;
 
   /**
-   * Enables cost confirmation via elicitation inside `create_project` for
-   * clients that declare per-request form-elicitation capability. Clients
-   * without that capability keep using `get_cost` -> `confirm_cost` ->
+   * Enables cost confirmation via elicitation for clients that declare
+   * per-request form-elicitation capability. Clients without that
+   * capability keep using `get_cost` -> `confirm_cost` ->
    * `create_project(confirm_cost_id)`.
    */
-  projectCostConfirmation?: {
+  costConfirmation?: {
     /** HMAC key for the `requestState` codec. MUST be at least 32 bytes. */
     requestStateKey: string | Uint8Array;
     /** The authenticated principal `requestState` is bound to. */
     principal: string;
     /** How long a minted `requestState` stays valid, in seconds. */
     ttlSeconds?: number;
+    /** Tools that accept a cost-confirmation elicitation. */
+    enabledTools: readonly 'create_project'[];
   };
 };
 
@@ -115,7 +115,7 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
     features,
     contentApiUrl = 'https://supabase.com/docs/api/graphql',
     onToolCall,
-    projectCostConfirmation,
+    costConfirmation,
   } = options;
 
   const contentApiClientPromise = createContentApiClient(contentApiUrl, {
@@ -135,12 +135,13 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
     features ?? availableDefaultFeatures
   );
 
-  const projectCostConfirmationCodec = projectCostConfirmation
-    ? createRequestStateCodec<ProjectCostState>({
-        key: projectCostConfirmation.requestStateKey,
-        ttlSeconds: projectCostConfirmation.ttlSeconds,
-        bind: (ctx) =>
-          `${ctx.mcpReq.method}:${projectCostConfirmation.principal}`,
+  const costConfirmationCodec = costConfirmation?.enabledTools.includes(
+    'create_project'
+  )
+    ? createRequestStateCodec<CostConfirmationState>({
+        key: costConfirmation.requestStateKey,
+        ttlSeconds: costConfirmation.ttlSeconds,
+        bind: (ctx) => `${ctx.mcpReq.method}:${costConfirmation.principal}`,
       })
     : undefined;
 
@@ -163,8 +164,8 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
       ]);
     },
     onToolCall,
-    requestState: projectCostConfirmationCodec && {
-      verify: projectCostConfirmationCodec.verify,
+    requestState: costConfirmationCodec && {
+      verify: costConfirmationCodec.verify,
     },
     tools: async () => {
       const contentApiClient = await contentApiClientPromise;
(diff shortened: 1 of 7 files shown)


packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -262,11 +297,135 @@ export function getAccountTools({ account, readOnly }: AccountToolsOptions) {
...
+            state.cost.amount !== cost.amount
+          ) {
+            // Pricing changed since the state was minted (e.g. the org's
+            // plan or active-project count shifted) - reissue a fresh
+            // prompt bound to the recomputed cost rather than honoring a
+            // stale quote.
+            return askForConfirmation();
+          }
+
+          const response = inputResponse(
```
Rodriguespn, line 383: I understand that `inputResponse` is defined by the official sdk and in theory, `response.kind` would always be a bounded and predictable value but I would prefer that the default behavior was either `cancel` or `decline` (imo cancel makes more sense to be the default but I'll leave it to your own judgement here), instead of defaulting to `accept`. 
Just trying to prevent if any unexpected behavior is introduced (for example, a new response type introduced in a later version) we wouldn't create the project

barryroodt: Good catch, thanks! 

Fixed

packages/mcp-utils/src/server.ts
```diff
@@ -540,6 +575,22 @@ export function createMcpServer(options: McpServerOptions) {
...
+    value !== null &&
+    (value as { resultType?: unknown }).resultType === 'input_required'
+  );
+}
+
+function isCallToolResult(value: unknown): value is CallToolResult {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    Array.isArray((value as { content?: unknown }).content)
```
Rodriguespn: Any tool returning { content: [...] } as ordinary data gets passed through raw here. I tried it with a { title, content: [...] } payload and the client rejected the whole result (`Invalid tools/call result: invalid_union`). The data never arrives. 
Since this is published and shared, it hits every consumer. Could we use an explicit discriminant like `isInputRequiredResult` [does](https://github.com/supabase/mcp/pull/391/changes#diff-f19f8fb1f7ad7fe85fd47d49a75422d3911fff89c0d0b470eda66307c25b537dR580), rather than the shape check?

packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -262,11 +297,135 @@ export function getAccountTools({ account, readOnly }: AccountToolsOptions) {
...
+                  type: 'text' as const,
+                  text: 'Request state arguments do not match the current arguments.',
+                },
+              ],
+              structuredContent: { status: 'error' },
+              isError: true,
+            };
+          }
+
+          if (
```
Rodriguespn: This drift check sits above the decline/cancel checks, so a decline gets dropped and re-prompted whenever the quote moved between rounds. Confirmed this locally: declining after a price shift returns `input_required` instead of `{status:'declined'}`. 
Moving the two action checks above this block should do it.

packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -262,11 +297,135 @@ export function getAccountTools({ account, readOnly }: AccountToolsOptions) {
...
+          const { codec } = costConfirmation;
+          const cost = await getNextProjectCost(account, organization_id);
+
+          const askForConfirmation = async () =>
+            inputRequired({
+              inputRequests: {
+                confirm_cost: inputRequired.elicit({
+                  mode: 'form',
+                  message: [
+                    `Creating this project costs $${cost.amount}/month.`,
```
Rodriguespn: `/month` is hardcoded here (and on line 330), but `cost.recurrence` already tells us 'monthly' vs 'hourly'. Correct today since only `create_project` is enabled, but nothing type-checks the string, so whenever `enabledTools` widens to `create_branch`, this will quietly render an hourly amount as `/month`. Cheap to read it from [`cost.recurrence`](https://github.com/supabase/mcp/pull/391/changes#diff-7bc305aaf1d0d8becf193477c0793d256c689c64549301c3c93563fb2e77562dR380) now.

packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -262,11 +297,135 @@ export function getAccountTools({ account, readOnly }: AccountToolsOptions) {
...
+                {
+                  type: 'text' as const,
+                  text: 'Project creation was cancelled.',
+                },
+              ],
+              structuredContent: { status: 'cancelled' },
+            };
+          }
+
+          return await account.createProject({
```
Rodriguespn, line 428: Replaying the same `requestState + accept` creates a project each time: The user gets 2 projects from one consent locally. I see replay prevention is called out as out of scope, so mostly wanted to check that's a deliberate call for Stage A given it's a billable action, and whether the platform PR handles it.

packages/mcp-server-supabase/src/tools/account-tools.ts
```diff
@@ -262,11 +297,135 @@ export function getAccountTools({ account, readOnly }: AccountToolsOptions) {
...
+          organization_id,
+          confirm_cost_id,
+        }: z.infer<typeof createProjectInputSchemaWithElicitation>,
+        ctx: ServerContext
+      ) => {
         if (readOnly) {
           throw new Error('Cannot create a project in read-only mode.');
         }
 
+        if (costConfirmation && isFormCapable(ctx)) {
```
Rodriguespn, line 316: We elicit even when the cost is $0, so free-tier creates get a round-trip and a "costs $0/month" prompt. Matches the legacy flow's behaviour, so no regression, but since the point here is smoothing the UX, is skipping the prompt at `amount === 0` on the table, or is confirming-regardless intentional?

packages/mcp-server-supabase/src/tools/cost-confirmation.ts
```diff
@@ -0,0 +1,59 @@
...
+  const capabilities = envelope[CLIENT_CAPABILITIES_META_KEY] as
+    | { elicitation?: Record<string, unknown> }
+    | undefined;
+  const elicitation = capabilities?.elicitation;
+  if (elicitation === undefined) {
+    return false;
+  }
+
+  const modes = Object.keys(elicitation);
+  return modes.length === 0 || modes.includes('form');
```
Rodriguespn, line 58: Checked this against the `2026-07-28` spec and it's exactly right ([spec](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation#capabilities)):

>  For backwards compatibility, an empty capabilities object is equivalent to declaring support for form mode only:
>   "elicitation": {}, // Equivalent to { "form": {} }

> Clients declaring the elicitation capability MUST support at least one mode (form or url).

> Servers MUST NOT send elicitation requests with modes that are not supported by the client.
 

An empty map is spec'd as `{ form: {} }`.
 
Only footnote: the reference client's `getSupportedElicitationModes` treats any map without form/url as form-capable, so `{ someFutureMode: {} }` differs here. We fall back to the legacy path, which is the safer side, and such a client is non-conforming anyway. Fine as-is IMO.

Rodriguespn (changes requested):
Left some comments

packages/mcp-utils/src/server.ts
barryroodt: good catch, fixed.

packages/mcp-server-supabase/src/tools/account-tools.ts
barryroodt: fixed

packages/mcp-server-supabase/src/tools/account-tools.ts
barryroodt: nice! fixed.

packages/mcp-server-supabase/src/tools/account-tools.ts
barryroodt: So yes, this is a deliberate call and won't be part of this initial deployment.

Adding replay prevention requires introducing state so that we can track use across istances - something the SDK does not cater to out of the box.

It's something we can look at adding as a follow-up when Platform have capacity available to advise.

packages/mcp-server-supabase/src/tools/account-tools.ts
barryroodt: I debated whether to cater for this or not (had it in previous iterations) - decided to keep things as simple as possible and leave it for follow-up work. In hindsight, I should have just handled it from the outset.

Fixed.

packages/mcp-server-supabase/src/tools/cost-confirmation.ts
barryroodt: Thanks for double checking this. Leaving as is for now.

packages/mcp-server-supabase/src/tools/account-tools.ts
Rodriguespn: Just note that I'm not fully across this area, so this is my understanding rather than a confident recommendation. Feel free to correct me where I've got it wrong.

Your mechanism claim looks right to me. The SDK's minted envelope is [`{ p, exp, b? }`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/requestStateCodec.ts#L238-L244), so there's no `jti` to track without adding one. And the [2026-07-28 spec](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr#server-requirements-basic-workflow) says:

> Note that these measures bound the replay window and prevent cross-user and cross-request reuse, but do not by themselves guarantee single-use. Servers for which a given `requestState` must be consumed at most once (e.g., one-time redemptions) **MUST** enforce that invariant server-side.

One correction to my original comment. I ran your branch against the local mgmt-api and replayed one accepted `requestState` three times: the first creates the project, the next two return `Project with name "…" already exists in your organization.`, and the row count stays at 1. That's the duplicate-name check in [`runProjectCreationFlow`](https://github.com/supabase/platform/blob/develop/packages/api-core/src/lib/projects/project-creation.ts#L574-L597). The two projects I saw before came from the MCP repo's MSW mock, which has no uniqueness rules. 

**Is it worth noting in the PR description or review the MSW mocks to try to reproduce this behavior?**

---

What I'm less sure about is the exemptions on that check (`AI_BUILDER_ORG_IDS`, `PLATFORM` tier, `highProjectCardinalityOrganizations`, `CONTINUOUS_TESTING_ORG_IDS`), since those are the orgs creating projects programmatically.

On the store: is cross-instance state actually needed here? My reading is that `requestState` is already unique per consent round and the hosted route has it on `req.body`, so `sha256(requestState)` would give a key every instance derives identically with no coordination, stored in the `metadata` jsonb `runProjectCreationFlow` already accepts, with a partial unique index on `(organization_id, metadata->>'idempotency_key')` for concurrent redemptions. 

I tried it locally to sanity-check the shape (small change of ~50 lines, no schema migration, no MCP package change) and the replays returned the same project ref instead of a 400.

If that holds, it seems preferable mainly because it removes the thing currently blocking this. There'd be nothing for Platform to advise on, no new stateful dependency, and no fail-open/fail-closed decision when a KV is unreachable. It'd also match how `v1-apply-a-migration` already does idempotency. But you know this code far better than I do, so I may well be missing a reason it doesn't work.

PS: keying off `requestState` wouldn't cover the `$0` path, since there's no elicitation and no state there.

Rodriguespn (approved):
Tested locally one more time and everything seems ok. Left a non-blocking comment with. suggestion to reduce the dependency of platform on duplicated ellicitation requests - https://github.com/supabase/mcp/pull/391#discussion_r3905850759

packages/mcp-server-supabase/src/server.ts
```diff
@@ -54,6 +56,23 @@ export type SupabaseMcpServerOptions = {
...
+  costConfirmation?: {
+    /** HMAC key for the `requestState` codec. MUST be at least 32 bytes. */
+    requestStateKey: string | Uint8Array;
+    /** The authenticated principal `requestState` is bound to. */
+    principal: string;
+    /** How long a minted `requestState` stays valid, in seconds. */
+    ttlSeconds?: number;
+    /** Tools that accept a cost-confirmation elicitation. */
+    enabledTools: readonly 'create_project'[];
+  };
```
gregnr, line 75: Is it generally best practice to create a single key/principle/ttl/etc per elicitation, or just set these once to be re-used by all elicitations?

gregnr: Other question - what are the pros of defining each of these properties ourselves vs simply accepting a `RequestStateCodec` that the upstream consumer builds themselves?

(I very likely don't have the full picture of how codecs are used btw - e.g. if codec instances are per-request or bound to a tool, my above points don't make sense)

gregnr (commented):
Very clean, nice 👍

packages/mcp-server-supabase/src/server.ts
barryroodt: > Is it generally best practice to create a single key/principle/ttl/etc per elicitation, or just set these once to be re-used by all elicitations?

yeah I fully expect this to be refactored once we start introducing further elicitations. We'll have to consider for instance whether we want different ttls for form vs url elicitations for example.

> Other question - what are the pros of defining each of these properties ourselves vs simply accepting a RequestStateCodec that the upstream consumer builds themselves?

No you have the right idea and it's a great question; original intention was to keep Platform modifications to a minimum, thus moving the api boundary out into the package.

For now I'd keep it as is. It keeps the internal state type and codec setup out of Platform. The only thing we'd be gaining is a smaller interface here.

If however, we want to introduce key rotation or some other codec implementation with security-sensitive binding logic, we can move it out into Platform without issue.

Thanks for the feedback, appreciate it.

gregnr (approved):

Issue #285, opened 2026-05-15
create_branch uncallable in project-scoped mode — confirm_cost not exposed


ocean-views-228:
# Bug report

## Describe the bug

When using the Supabase MCP server in project-scoped mode (recommended security configuration), the branching tools are fully non-functional:

1. `create_branch` requires a `confirm_cost_id` parameter, which can only be obtained from `confirm_cost`. But `confirm_cost` (and `get_cost`) are in the Account Management feature group, which is disabled in project-scoped mode. This makes `create_branch` uncallable.

2. `list_branches` returns the error: "Project reference is missing when validating permissions." This makes the entire branching surface non-functional — even if a branch were created via the dashboard, the MCP cannot see or manage it.`merge_branch` is presumed to fail the same way.

The branching tools are listed in the Branching feature group, which is documented as available in project-scoped mode. The bug
is that they depend on an account-scoped prerequisite (`confirm_cost`) that is unavailable in the same mode.

## To Reproduce

1. Connect the Supabase MCP server in project-scoped mode (passing `project_ref` parameter) 
2. Call `list_branches` → error: "Project reference is missing when validating permissions"
3. Attempt to call `create_branch` → cannot provide required `confirm_cost_id` parameter because `confirm_cost` tool is
   not available in project-scoped mode
4. All branching operations are blocked

## Expected behavior

Branching tools should be fully functional in project-scoped mode. Either:
- Expose `get_cost` / `confirm_cost` in project-scoped mode (scoped to the connected project), or
- Allow `create_branch` without `confirm_cost_id` in project-scoped mode (project is already identified, cost is deterministic), or
- Fix project reference resolution for `list_branches` / `merge_branch` in project-scoped connections

## Screenshots

N/A — MCP tool calls, not browser UI.

## System information

- OS: N/A (cloud-hosted MCP via Claude.ai)
- Browser: N/A
- Version of supabase-js: N/A
- Version of Node.js: N/A
- MCP access method: Claude.ai MCP integration
- Supabase plan: Pro
- MCP connection mode: project-scoped
- Region: us-west-2

## Additional context

This blocks the staging-branch workflow for MCP-managed migrations. The intended workflow is create_branch → apply_migration on dev branch → verify → merge_branch to production. Without functional branching, migrations must be applied directly to production with no staging validation.

kayalopez:
This looks like a good regression target for project-scoped MCP mode because the broken path is not just `create_branch`; it is the whole migration-staging contract.

[...]

PR #394, opened 2026-08-29
feat(mcp): add branch cost confirmation elicitation

barryroodt:
Stacked on [#391](https://github.com/supabase/mcp/pull/391). Review that first. This is part of [AI-1091](https://linear.app/supabase/issue/AI-1091/ship-form-mode-cost-confirmation-stage-a-supabasemcp).

Feature: add form-mode cost confirmation to `create_branch`.

#391 provides the shared confirmation codec/verifier, `ServerContext`, result passthrough, and project flow. Branch creation still uses the existing hash validation.

## What is the new behavior?

Modern form clients can call `create_branch` without `confirm_cost_id`. They receive an action-only, property-less prompt for **$0.01344/hr** (about **$9.68 over 30 days**).

Acceptance creates the branch from signed arguments. Missing responses and changed costs re-prompt. A wrong-tool state or changed arguments fail. Decline and cancel create no branch.

Capability-free and URL-only clients keep hash validation. A supplied legacy ID cannot bypass the capable-client flow. Project-scoped injection works, and existing output fields such as `with_data` remain.

## How to Review

1. **Server and shared state**
   - `packages/mcp-server-supabase/src/server.ts`
   - `packages/mcp-server-supabase/src/tools/cost-confirmation.ts`

2. **Branch handler and pricing**
   - `packages/mcp-server-supabase/src/tools/branching-tools.ts`
   - `packages/mcp-server-supabase/src/pricing.ts`

3. **Behavioral coverage**
   - `packages/mcp-server-supabase/src/server.test.ts`
   - Inspect the capable, legacy, argument-binding, and project-scoped paths.

- [ ] Is `create_branch` enabled only through the intended server path?
- [ ] Does the discriminated state bind project ID, branch name, and `BranchCost`?
- [ ] Do capable clients always use elicitation, even with a legacy ID?
- [ ] Do accept, decline, cancellation, wrong-tool, and argument-change outcomes look correct?

## Verification

- `pnpm --filter @supabase/mcp-server-supabase test src/server.test.ts -t "create_(project|branch) cost confirmation via elicitation"` (17 passed)
- `pnpm --filter @supabase/mcp-server-supabase build` (typecheck + build passed)
- `pnpm test:packed-platform-consumer` (3 assertions passed)
- `pnpm exec biome ci .` (passed)

## Additional context

Commit `24535f1 chore: refresh management API types` adds two generated lines to `packages/mcp-server-supabase/src/management-api/types.ts` because the CI generation gate detected upstream schema drift. Feature behavior and verification remain unchanged.

This does not change `create_project`, `mcp-utils`, platform adapters, stdio, pricing values, manifests, or docs.

Known limits: confirmations are not single-use, legacy IDs remain precomputable, and the branch rate remains the existing package constant.

AI-assisted; human review required.

### DIFF
--- a/packages/mcp-server-supabase/src/pricing.ts
+++ b/packages/mcp-server-supabase/src/pricing.ts
@@ -48,6 +48,6 @@ export async function getNextProjectCost(
 /**
  * Gets the cost for a database branch.
  */
-export function getBranchCost(): Cost {
+export function getBranchCost(): BranchCost {
   return { type: 'branch', recurrence: 'hourly', amount: BRANCH_COST_HOURLY };
 }
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -3920,6 +3920,53 @@ describe('tools', () => {
       expect(mockBranches.size).toBe(0);
     });
 
+    test('form-capable client: a non-elicitation response re-prompts without creating a branch', async () => {
+      const { client } = await setupFormCapable();
+
+      const org = await createOrganization({
+        name: 'My Org',
+        plan: 'free',
+        allowed_release_channels: ['ga'],
+      });
+      const project = await createProject({
+        name: 'Project 1',
+        region: 'us-east-1',
+        organization_id: org.id,
+      });
+      project.status = 'ACTIVE_HEALTHY';
+
+      const args = { project_id: project.id, name: 'test-branch' };
+      const first = (await client.request(
+        {
+          method: 'tools/call',
+          params: { name: 'create_branch', arguments: args },
+        },
+        { allowInputRequired: true }
+      )) as CallToolResult | InputRequiredResult;
+
+      if (!isInputRequiredResult(first)) {
+        throw new Error('expected an input_required result');
+      }
+
+      const second = (await client.request(
+        {
+          method: 'tools/call',
+          params: {
+            name: 'create_branch',
+            arguments: args,
+            inputResponses: {
+              confirm_cost: { roots: [] },
+            },
+            requestState: first.requestState,
+          },
+        },
+        { allowInputRequired: true }
+      )) as CallToolResult | InputRequiredResult;
+
+      expect(isInputRequiredResult(second)).toBe(true);
+      expect(mockBranches.size).toBe(0);
+    });
+
     test('a decline is honored even when the quoted branch cost changed since the state was minted', async () => {
       const getBranchCostSpy = vi
         .spyOn(pricing, 'getBranchCost')
@@ -3959,9 +4006,12 @@ describe('tools', () => {
       if (!isInputRequiredResult(first)) {
         throw new Error('expected an input_required result');
       }
-      expect(first.inputRequests.confirm_cost).toMatchObject({
-        mode: 'form',
-        message: expect.stringContaining(`$${BRANCH_COST_HOURLY}/hr`),
+      expect(first.inputRequests?.confirm_cost).toMatchObject({
+        method: 'elicitation/create',
+        params: {
+          mode: 'form',
+          message: expect.stringContaining(`$${BRANCH_COST_HOURLY}/hr`),
+        },
       });
 
       const second = (await client.request(
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -4202,6 +4202,134 @@ describe('tools', () => {
         )
       ).toHaveLength(1);
     });
+
+    test('rejects a requestState minted by create_project', async () => {
+      const { client } = await setupFormCapable();
+
+      // create_project only triggers cost confirmation for a paid org's
+      // additional projects, so set up a pro org with one existing project.
+      const org = await createOrganization({
+        name: 'Paid Org',
+        plan: 'pro',
+        allowed_release_channels: ['ga'],
+      });
+      const existingProject = await createProject({
+        name: 'Existing Project',
+        region: 'us-east-1',
+        organization_id: org.id,
+      });
+      existingProject.status = 'ACTIVE_HEALTHY';
+
+      const projectFirst = (await client.request(
+        {
+          method: 'tools/call',
+          params: {
+            name: 'create_project',
+            arguments: {
+              organization_id: org.id,
+              name: 'My Project',
+              region: 'us-east-1',
+            },
+          },
+        },
+        { allowInputRequired: true }
+      )) as CallToolResult | InputRequiredResult;
+
+      if (!isInputRequiredResult(projectFirst)) {
+        throw new Error(
+          'expected an input_required result from create_project'
+        );
+      }
+
+      const result = (await client.request(
+        {
+          method: 'tools/call',
+          params: {
+            name: 'create_branch',
+            arguments: { project_id: existingProject.id, name: 'test-branch' },
+            inputResponses: {
+              confirm_cost: { action: 'accept', content: {} },
+            },
+            requestState: projectFirst.requestState,
+          },
+        },
+        { allowInputRequired: true }
+      )) as CallToolResult | InputRequiredResult;
+
+      if (isInputRequiredResult(result)) {
+        throw new Error('expected a CallToolResult');
+      }
+
+      expect(result.isError).toBe(true);
+      expect(result.structuredContent).toEqual({ status: 'error' });
+      expect(
+        result.content.some(
+          (c) =>
+            c.type === 'text' &&
+            c.text === 'Request state was not issued for create_branch.'
+        )
+      ).toBe(true);
+      expect(mockBranches.size).toBe(0);
+    });
+
+    test('rejects a tampered requestState before the handler runs', async () => {
+      const { client } = await setupFormCapable();
+
+      const org = await createOrganization({
+        name: 'My Org',
+        plan: 'free',
+        allowed_release_channels: ['ga'],
+      });
+      const project = await createProject({
+        name: 'Project 1',
+        region: 'us-east-1',
+        organization_id: org.id,
+      });
+      project.status = 'ACTIVE_HEALTHY';
+
+      const first = (await client.request(
+        {
+          method: 'tools/call',
+          params: {
+            name: 'create_branch',
+            arguments: { project_id: project.id, name: 'test-branch' },
+          },
+        },
+        { allowInputRequired: true }
+      )) as CallToolResult | InputRequiredResult;
+
+      if (!isInputRequiredResult(first)) {
+        throw new Error('expected an input_required result');
+      }
+
+      // Flip the last character to tamper the HMAC signature; the framework
+      // rejects it before the handler runs (ProtocolError -32602).
+      const originalState = first.requestState as string;
+      const lastChar = originalState.slice(-1);
+      const tamperedState =
+        originalState.slice(0, -1) + (lastChar === 'a' ? 'b' : 'a');
+
+      await expect(
+        client.request(
+          {
+            method: 'tools/call',
+            params: {
+              name: 'create_branch',
+              arguments: { project_id: project.id, name: 'test-branch' },
+              inputResponses: {
+                confirm_cost: { action: 'accept', content: {} },
+              },
+              requestState: tamperedState,
+            },
+          },
+          { allowInputRequired: true }
+        )
+      ).rejects.toMatchObject({
+        code: -32602,
+        message: 'Invalid or expired requestState',
+      });
+      expect(mockBranches.size).toBe(0);
+    });
   });
 
   test('delete branch', async () => {
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -3980,61 +3980,64 @@ describe('tools', () => {
           recurrence: 'hourly',
           amount: BRANCH_COST_HOURLY + 1,
         });
-      const { client } = await setupFormCapable();
+      try {
+        const { client } = await setupFormCapable();
 
-      const org = await createOrganization({
-        name: 'My Org',
-        plan: 'free',
-        allowed_release_channels: ['ga'],
-      });
-      const project = await createProject({
-        name: 'Project 1',
-        region: 'us-east-1',
-        organization_id: org.id,
-      });
-      project.status = 'ACTIVE_HEALTHY';
+        const org = await createOrganization({
+          name: 'My Org',
+          plan: 'free',
+          allowed_release_channels: ['ga'],
+        });
+        const project = await createProject({
+          name: 'Project 1',
+          region: 'us-east-1',
+          organization_id: org.id,
+        });
+        project.status = 'ACTIVE_HEALTHY';
 
-      const args = { project_id: project.id, name: 'test-branch' };
-      const first = (await client.request(
-        {
-          method: 'tools/call',
-          params: { name: 'create_branch', arguments: args },
-        },
-        { allowInputRequired: true }
-      )) as CallToolResult | InputRequiredResult;
-
-      if (!isInputRequiredResult(first)) {
-        throw new Error('expected an input_required result');
-      }
-      expect(first.inputRequests?.confirm_cost).toMatchObject({
-        method: 'elicitation/create',
-        params: {
-          mode: 'form',
-          message: expect.stringContaining(`$${BRANCH_COST_HOURLY}/hr`),
-        },
-      });
-
-      const second = (await client.request(
-        {
-          method: 'tools/call',
-          params: {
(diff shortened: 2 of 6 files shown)


packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -156,7 +164,7 @@ const COST_CONFIRMATION: NonNullable<
 > = {
   requestStateKey: 'a'.repeat(32),
   principal: 'test-user',
-  enabledTools: ['create_project'],
+  enabledTools: ['create_project', 'create_branch'],
```
Rodriguespn, line 167: nit: `enabledTools` is both tools everywhere in the suite, but platform #37717 ships `['create_project']`. I checked both single-tool configs by hand and the wiring is correct. Just nothing pinning it, so an inverted `includes()` at server.ts:195/237 would go unnoticed. Worth a small case per config.

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -155,18 +184,137 @@ export function getBranchingTools({
...
+                ctx
+              ),
+            });
+
+          const state = ctx.mcpReq.requestState<CostConfirmationState>();
+          if (!state) {
+            return askForConfirmation();
+          }
+
+          if (state.tool !== 'create_branch') {
```
Rodriguespn, line 239: nit: This branch isn't covered. I minted a `create_project` state and sent it here (and a branch state to `create_project`), both correctly return `Request state was not issued for …` and create nothing, so the behaviour is right. Since #394 is the first PR where two tools share one codec, this feels worth pinning.

packages/mcp-server-supabase/src/server.ts
Rodriguespn, line 143: nit: `method` is `tools/call` for both tools, so `bind` doesn't separate them. The payload `tool` field is the only thing that does. That's checked first in both handlers so it's fine, but including the tool name here would let the codec reject a cross-tool state before the handler runs.

packages/mcp-server-supabase/src/pricing.ts
Rodriguespn, line 26: nit: You tightened `getBranchCost()` to `BranchCost`,  worth doing the same here (`Promise<ProjectCost>`) and at `cost-confirmation.ts:19`, where `ProjectCostState.cost` is still the wide `Cost`. Harmless today since the drift check compares type, recurrence and amount, but this is the PR introducing the discrimination.

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -155,18 +184,137 @@ export function getBranchingTools({
...
+              inputRequests: {
+                confirm_cost: inputRequired.elicit({
+                  mode: 'form',
+                  message: [
+                    `Creating this branch costs $${cost.amount}${costSuffix} while it runs.`,
+                    '',
+                    `Project          ${project_id}`,
+                    `Branch           ${name}`,
+                    `Hourly rate      $${cost.amount}${costSuffix}`,
+                    `30-day estimate  ~$${(cost.amount * 24 * 30).toFixed(2)} if left running`,
```
Rodriguespn: Question rather than a request: project cost is derived from the org's plan, but `getBranchCost()` is a flat constant with no org awareness, and platform has org-level branching entitlements plus a `FREE_BRANCHING_ORG_IDS` exemption. Does `~$9.68 over 30 days` hold for every org, or can it overstate what some are charged? Happy to be told it's fine.

Rodriguespn (commented):
This does read as #391 with the branch flow swapped in, and I can confirm the fixes from my earlier review all carried over (safe defaults on the elicitation action, drift check below decline/cancel, and the suffix driven off `cost.recurrence` rather than hardcoded).

Ran the elicitation suite locally (21 passing) plus a few probes of my own for the paths the suite doesn't cover, both cross-tool state and the single-tool `enabledTools` configs. Both behave correctly; I've left inline comments suggesting tests for them, plus a couple of small type/pricing notes. Nothing blocking.

One thing worth adding to the description: replay is better contained here than for projects. `preview_branches` has a real `UNIQUE (project_id, branch_name)` constraint ([migration](https://github.com/supabase/platform/blob/develop/middleware-db/migrations/20250618145142_add_unique_branch_name_per_project.sql)), so a replayed branch confirmation fails at the DB rather than relying on the read-then-check that covers `create_project`. Same known limit, weaker consequence.

packages/mcp-server-supabase/src/server.test.ts
barryroodt: Yeah, that would add useful coverage for the single-tool configuration. I’m leaving it out of this PR to keep the change focused on the shared request-state flow.

packages/mcp-server-supabase/src/tools/branching-tools.ts
barryroodt: Thanks,  I’m leaving it out of this PR; the current handlers already reject state signed for the other tool.

packages/mcp-server-supabase/src/server.ts
barryroodt: No change needed here. The SDK bind context exposes tools/call, not the tool name, while the signed payload includes tool and each handler rejects a mismatch.

packages/mcp-server-supabase/src/pricing.ts
barryroodt: I’m leaving the `ProjectCost` narrowing unchanged here. It’s compile-time-only and would modify types owned by parent PR #391 from within child PR #394.

packages/mcp-server-supabase/src/tools/branching-tools.ts
barryroodt: Good catch. Fixed.

The prompt now says rates and estimates are standard before plan allowances or exemptions, and that actual charges may be lower.

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -155,18 +184,137 @@ export function getBranchingTools({
...
+          const cost = getBranchCost();
+          const costSuffix = { hourly: '/hr' }[cost.recurrence];
+
+          const askForConfirmation = async () =>
+            inputRequired({
+              inputRequests: {
+                confirm_cost: inputRequired.elicit({
+                  mode: 'form',
+                  message: [
+                    `Creating this branch costs $${cost.amount}${costSuffix} while it runs.`,
```
mattrossman: This wording is a bit unclear to me:
- Is the cost coming from the time it takes to create the branch, or any time that the branch is alive?
- What determines if a branch is running? Do branches stop on their own? [Docs mention](https://supabase.com/docs/guides/deployment/branching/troubleshooting#connection-timeouts) that preview branches can auto-pause but it may not be obvious to someone using this tool w/o that context
- What kind of branch does this tool create? Persistent or preview? [Docs describe](https://supabase.com/docs/guides/deployment/branching#how-branching-works) different lifecycle for each.

I believe our `create_branch` tool only creates preview branches.

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -155,18 +184,137 @@ export function getBranchingTools({
...
+                    '',
+                    `Project          ${project_id}`,
+                    `Branch           ${name}`,
```
mattrossman: Something to be aware of — in Claude Code `2.1.258` launched with `MCP_SDK_GENERATION=v2 MCP_PROTOCOL_NEGOTIATION=auto claude`, I can't actually read any of the elicitation prompt from here onwards, it's collapsed and I don't know if there's a way to expand it. The `Ctrl + o` shortcut I'd otherwise use to expand transcripts doesn't seem to do it. We may want to consider shortening the elicitation to make sure all important details are visible.

https://github.com/user-attachments/assets/8b07bbf0-9cb2-4623-95ab-8df36451ef47

packages/mcp-server-supabase/src/tools/branching-tools.ts
```diff
@@ -155,18 +184,137 @@ export function getBranchingTools({
...
+                  requestedSchema: actionOnlyElicitationSchema,
+                }),
+              },
+              requestState: await codec.mint(
+                { tool: 'create_branch', project_id, name, cost },
+                ctx
+              ),
+            });
+
+          const state = ctx.mcpReq.requestState<CostConfirmationState>();
```
mattrossman, line 234: I assume it's safe to trust this type w/o runtime validation? Even though we prevent state tampering I wondering if there is any reason we could receive a shape we don't expect here.

packages/mcp-server-supabase/src/tools/branching-tools.ts
barryroodt: Preview only, billed hourly for as long as the branch exists, and yes, preview branches auto-pause on inactivity. The handler posts `{ branch_name: name }` with no `persistent` field ([api-platform.ts L680-682](https://github.com/supabase/mcp/blob/2370f114e0972a36f5c3895ae7bf52987ad776ff/packages/mcp-server-supabase/src/platform/api-platform.ts#L680-L682)), so the API default applies. Whether a paused branch still accrues isn't stated in the [branching usage docs](https://supabase.com/docs/guides/platform/manage-your-usage/branching) (they count hours the branch existed), I'll confirm with billing. Until then I'd word it as 'until deleted' so we never understate, and add the lifecycle facts to the prompt:

```
Preview branch: $0.01344/hr until deleted (~$9.68 per 30 days).
Auto-pauses on inactivity. https://supabase.com/docs/guides/deployment/branching#how-branching-works
Standard rate, before plan allowances or exemptions.
```

packages/mcp-server-supabase/src/tools/branching-tools.ts
barryroodt: Good catch. I can't reproduce it here, but going by your description the "Cost recurs until deletion" line sits in the hidden tail, and that's the one line that has to be visible. The rewrite I'm proposing in the wording thread fixes this too (3 short lines, the important details in the first two):

```
Preview branch: $0.01344/hr until deleted (~$9.68 per 30 days).
Auto-pauses on inactivity. https://supabase.com/docs/guides/deployment/branching#how-branching-works
Standard rate, before plan allowances or exemptions.
```

`create_project` in #391 has the same structure and the same collapse point. I'll fix `create_branch` here in #394 and amend #391 for `create_project`, it's a small text change and #391 hasn't merged yet.

packages/mcp-server-supabase/src/tools/branching-tools.ts
mattrossman, line 234: Non-blocking, this section (checking that state exists, checking the tool name + other args match, decline/cancel, price-change re-prompt) is the same as `create_project`'s, maybe worth making a shared/tested helper as I imagine most tools will need this.

I had a similar question as what Pedro raised [here](https://github.com/supabase/mcp/pull/394/changes#r3906180326) about us needing to manually verify the tool name instead of putting it in `bind`, and since the conclusion was that `bind` context doesn't see the tool name, the helper could also prevent mistakes there.

barryroodt: Yeah, it's safe against tampering: the codec HMAC-verifies before the handler ever sees the state, so anything we didn't sign is rejected. The one gap is version skew, where a token minted before a deploy still passes within the TTL with an old shape, and today the field checks (starting at [line 239](https://github.com/supabase/mcp/blob/07425eb34b32a7e13ad479a0939984e35dda04da/packages/mcp-server-supabase/src/tools/branching-tools.ts#L239)) catch that with a re-prompt or an error before any branch is created. I'd add a `safeParse` of the decoded state as a follow-up rather than in this PR.

barryroodt: Agree, a shared helper is the right follow-up.  I'd do it once #391 and #394 have merged, since #391 owns `create_project` and would need the same change.

packages/mcp-server-supabase/src/tools/branching-tools.ts
barryroodt: Pushed the 3-line prompt for create_branch (07425eb) and rebased onto the amended #391 head. Same shape applied to create_project in #391 (1d10c3c).

mattrossman: That's much better, thanks!

Annoyingly, in a narrow window (e.g. split panes in VS Code) Claude Code truncates the URL and clicking it opens the truncated version instead of expanding it. Not sure if there's anything we can do there, maybe drop the URL if it's not adding much beyond what the prompt already shows, or replacing one of the lines with it to reduce truncation likelihood

CleanShot 2026-09-02 at 23 11 42@2x

CleanShot 2026-09-02 at 23 11 46@2x

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -3732,6 +3742,468 @@ describe('tools', () => {
...
+        { allowInputRequired: true }
+      )) as CallToolResult | InputRequiredResult;
+
+      if (!isInputRequiredResult(result)) {
+        throw new Error('expected an input_required result');
+      }
+      expect(mockBranches.size).toBe(0);
+    });
+
+    test('rejects a retry whose arguments changed since the state was minted', async () => {
```
mattrossman, line 4102: Might be good to add a test that a state minted by `create_project` is rejected by `create_branch`

packages/mcp-server-supabase/src/server.test.ts
mattrossman, line 214: Could we test some other `capabilitites` shapes so that `isFormCapable` gets some coverage? Maybe it could be parameterized here.

For example I didn't know at first that `{ elicitations: {} }` is [equivalent to form elicitations](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation#capabilities) for backwards compat. Also could test clients that pass URL-only `{ elicitations: { url: {} } }` or empty capabilities `{}`

packages/mcp-server-supabase/src/server.ts
mattrossman, line 167: Not a huge risk, but it could be nice to have a test verifying that a tampered state is properly [rejected](https://github.com/modelcontextprotocol/typescript-sdk/blob/cc4b41617ce3601b1290d67216ea0b194a3cd9ac/packages/server/src/server/server.ts#L713), in case we accidentally mess up this `verify` wiring in the future.

mattrossman (approved):
Tested this by fusing changes from #402 onto the branch locally to connect to Claude Code launched w/ the modern protocol flags:

```bash
MCP_SDK_GENERATION=v2 MCP_PROTOCOL_NEGOTIATION=auto claude
```

This was my first time seeing elicitations working in Claude which was cool :) It was working as expected. Would love to get the HTTP dev server in so future stuff stays low lift to test.

Left a couple minor comments about test coverage opportunities. The Claude Code elicitation UI design is questionable w/ how it truncates info, but that's not something we have a lot control over and who knows how it'll change over time. Leaving it to you how you want to address the remaining feedback.

Noting this PR will partially address https://github.com/supabase/mcp/issues/285, in that the unavailable cost confirmation tools aren't needed w/ elicitations 🙌

packages/mcp-server-supabase/src/tools/branching-tools.ts
barryroodt: Dropped the URL from both prompts, the text already carried the fact: [c3e5ed8](https://github.com/supabase/mcp/pull/394/commits/c3e5ed8e7c6b9f41fd53be2a110da6a38ba86bd2) here and [171060b](https://github.com/supabase/mcp/pull/391/commits/171060be16ec866aaa99c0bd054dd6ab936a7e93) in #391.

packages/mcp-server-supabase/src/server.test.ts
barryroodt: Added in [07c889d](https://github.com/supabase/mcp/pull/394/commits/07c889d9f7170ed49b9b1ac8a9a6bf03593ae392): a `create_project` state sent to `create_branch` gets the 'not issued for create_branch' error and creates nothing.

packages/mcp-server-supabase/src/server.test.ts
barryroodt: Good shout, added in [07c889d](https://github.com/supabase/mcp/pull/394/commits/07c889d9f7170ed49b9b1ac8a9a6bf03593ae392) as a new `cost-confirmation.test.ts` covering five shapes: empty, `{ elicitation: {} }`, url-only, form, and a client with no protocol version.

packages/mcp-server-supabase/src/server.ts
barryroodt: Added in [07c889d](https://github.com/supabase/mcp/pull/394/commits/07c889d9f7170ed49b9b1ac8a9a6bf03593ae392): a tampered state is rejected before the handler runs with `-32602` 'Invalid or expired requestState'.

PR #407, opened 2026-09-03
fix: ensure correct escaping in advisor

staaldraad:

Bug fix

Table names were not escaped in rls advisor.

## What is the new behavior?

Escapes table names to ensure valid SQL in the suggested ALTER TABLE statement.

### DIFF
--- a/packages/mcp-server-supabase/src/advisories/rls-disabled.ts
+++ b/packages/mcp-server-supabase/src/advisories/rls-disabled.ts
@@ -34,6 +34,26 @@ const SYSTEM_SCHEMAS = new Set([
 ]);
 
 /**
+ * Quotes a Postgres identifier, doubling any embedded double quotes.
+ */
+function quoteIdentifier(identifier: string): string {
+  return `"${identifier.replace(/"/g, '""')}"`;
+}
+
+/**
+ * Quotes a `schema.table` name for safe use in generated SQL, quoting the
+ * schema and table portions independently.
+ */
+function quoteQualifiedName(name: string): string {
+  const dotIndex = name.indexOf('.');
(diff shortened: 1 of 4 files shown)


raulb (approved):
Thank you also for updating the  mgmt-api types 🙇‍♂️

packages/mcp-server-supabase/src/advisories/rls-disabled.ts
```diff
@@ -33,6 +33,26 @@ const SYSTEM_SCHEMAS = new Set([
...
+function quoteIdentifier(identifier: string): string {
+  return `"${identifier.replace(/"/g, '""')}"`;
+}
+
+/**
+ * Quotes a `schema.table` name for safe use in generated SQL, quoting the
+ * schema and table portions independently.
+ */
+function quoteQualifiedName(name: string): string {
+  const dotIndex = name.indexOf('.');
```
barryroodt, line 48: Splitting on the first dot loses the schema boundary, and the filter below has the same ambiguity. For `{ schema: "my.app", name: "users" }`, this generates `"my"."app.users"`; a schema such as `auth.legacy` is also treated as `auth` and hidden. Could this use the original `schema` and `name` fields for both filtering and SQL?

barryroodt (commented):
Nice work! The identifier escaping fix looks correct. One minor issue remains: joining `schema` and `name`, then splitting on the first dot, breaks remediation SQL for schemas such as `my.app`. Could we pass `{ schema, name, rls_enabled }` into `buildRlsDisabledAdvisory`, use the separate fields for filtering and SQL, and keep `${schema}.${name}` only for display?

PR #408, opened 2026-09-03
feat(mcp)!: confirm destructive SQL via elicitation and rename costConfirmation option

barryroodt:
## Why

Adds confirmation elicitations before destructive SQL, such as dropping a table 

Closes [AI-1186](https://linear.app/supabase/issue/AI-1186).

## What changed

- Renames `costConfirmation` → `confirmation`. `enabledTools` 
- Supports existing cost confirmations (`create_project`, `create_branch`) and SQL confirmations (`execute_sql`, `apply_migration`); 
- Local HTTP opts into both SQL tools. 
- No new dependencies.

## How to test
<details>
<summary>Expand for details</summary>

1. **Choose one disposable target and prepare your tools.** Use either:
   - **Staging:** API `https://api.supabase.green`, an existing staging personal access token (PAT), and a disposable staging project ref.
   - **Dedicated disposable production:** API `https://api.supabase.com`, an existing production PAT, and that disposable production project's ref.

   The server API URL, PAT, and project ref must all match your choice. **Never use customer, shared, or application data.**

   Check out this PR (`feat/destructive-sql-confirmation`, reviewed head `8a7e9de`) locally. Have the repository's Node LTS and pnpm 10 active ([tool setup](https://github.com/supabase/mcp/blob/a4f435165357d9ddaa7bf3e7b0fc4bbfcb9bc07d/CONTRIBUTING.md)), plus a current Claude Code version with MCP elicitation form support. Disable elicitation auto-answer hooks for this session.

2. **Build and start the local HTTP server in Terminal A.** Replace `/path/to/this-pr-checkout` with your checkout root: the folder containing `package.json` and `packages/`.

   Use your existing pnpm 10.33.2 executable: replace `PNPM10` below with its absolute path before running. The subshell puts it ahead of Corepack shims so nested commands use the same version, without changing your shell configuration. No `approve-builds` step is needed; `enable-pre-post-scripts` enables the package’s own prebuild step, not dependency build approval.

   ```sh
   cd /path/to/this-pr-checkout && (
     PNPM10=/absolute/path/to/existing/pnpm10/pnpm
     export PATH="$(dirname "$PNPM10"):$PATH"
     export npm_config_manage_package_manager_versions=false

     "$PNPM10" --config.manage-package-manager-versions=false install --frozen-lockfile &&
     "$PNPM10" --config.manage-package-manager-versions=false --config.enable-pre-post-scripts=true --filter '@supabase/mcp-server-supabase...' build
   )
   ```

   For **staging**, start:

   ```sh
   node packages/mcp-server-supabase/dist/cli.js --http --port 3111 --api-url https://api.supabase.green
   ```

   For **dedicated disposable production**, replace only `https://api.supabase.green` with `https://api.supabase.com` in that command. Leave Terminal A running; HTTP mode does not need a server-side PAT export.

   Reuse an existing server only if it runs this PR with your chosen API URL and port. Otherwise, use a free port and change `3111` in both the launch command and the config below. Do not stop an unrelated server.

3. **Create a private `.mcp.json` at the checkout root.** Open Terminal B using interactive **bash or zsh**, then enter the same folder:

   ```sh
   cd /path/to/this-pr-checkout
   git ls-files -- .mcp.json
   ```

   If that command lists `.mcp.json`, stop: an ignore rule cannot protect a tracked file. If the file already exists, do not overwrite it or add a duplicate registration. Privately reuse only a suitable entry matching the local endpoint, disposable target, `read_only=false`, database tools, and environment-backed PAT shown below. If incompatible, stop rather than modify, delete, or untrack it. When reusing an entry, use its actual server name and environment-variable names in the remaining steps.

   **For a new file**, first add this repository-local exclusion, not a tracked `.gitignore` change:

   ```sh
   printf '\n/.mcp.json\n' >> "$(git rev-parse --git-path info/exclude)"
   ```

   In a local text editor, create `.mcp.json` directly beside the root `package.json`, **not** inside `packages/mcp-server-supabase` or `~/.claude`. Paste this JSON, keeping both `${...}` placeholders literal. Never put a token in the file.

   ```json
   {
     "mcpServers": {
       "supabase-sql-review": {
         "type": "http",
         "url": "http://127.0.0.1:3111/mcp?project_ref=${SUPABASE_REVIEW_PROJECT_REF}&read_only=false&features=database",
         "headers": {
           "Authorization": "Bearer ${SUPABASE_REVIEW_PAT}"
         }
       }
     }
   }
   ```

   Check that the file is ignored:

   ```sh
   git check-ignore .mcp.json
   ```

   Expect `.mcp.json` as output; otherwise stop. Keep this config local and do not force-add or commit it. The local URL stays the same for either environment; the server's `--api-url` selects staging or production.

4. **Enter your existing PAT and launch Claude from Terminal B.** Run these commands one at a time, answering each prompt before continuing. Enter the PAT only at the silent input prompt, never as a shell command or JSON value. Stop if either input is missing or does not match your chosen environment.

   ```sh
   printf 'Chosen disposable project ref: ' >&2
   IFS= read -r SUPABASE_REVIEW_PROJECT_REF
   export SUPABASE_REVIEW_PROJECT_REF
   printf 'Existing PAT for the same environment: ' >&2
   IFS= read -r -s SUPABASE_REVIEW_PAT
   printf '\n' >&2
   export SUPABASE_REVIEW_PAT
   claude --tools "" --strict-mcp-config --mcp-config .mcp.json
   ```

   In Claude, use `/mcp` to confirm that `supabase-sql-review` (or your reused entry) is connected. Privately verify the actual target, not just the connection label. Stop on ambiguity or connection failure; do not switch servers or use an OAuth workaround.

   `--tools ""` disables built-in tools, not MCP tools. Strict config loads only the supplied config, but **every server in that file can load**. Use only the selected server's `execute_sql` and `apply_migration`. The PAT remains in process environment/memory: do not print the environment, expose secrets in logs/screenshots, or ask Claude to inspect the config.

5. **Copy the prompt below into Claude**, replacing its three bracketed header values. This is the direct-SQL confirmation matrix for #408, not the parser matrix for #426. Answer actual server elicitation forms manually. Normal tool-permission dialogs and initial target confirmation do not replace per-operation forms.

### Reusable #408 manual test prompt

```text
Target environment: [staging or dedicated disposable production]
MCP server: [supabase-sql-review or selected server]
Exact disposable project ref: [...]

Run this manual check only against the selected MCP server and the exact disposable project above. Confirm the target with me before any writes. Never use a customer, shared, or application-data project. Before starting, confirm that both execute_sql and apply_migration are available and the client supports form elicitation. Do not change or bypass configured readOnly, per-tool, or form gates. Stop if the target or prerequisites are ambiguous.

Use only that server's execute_sql and apply_migration tools. Do not use shell, HTTP, psql, other servers, configuration or secret reads, new credentials, server changes, automatic response hooks, or scripted approvals. I must answer every server elicitation form manually. A tool-permission dialog or approval in chat is not that form. If you cannot observe the form or my response, ask me what appeared; do not infer it from the tool result.

This is not a dry run. Setup, two accepted UPDATEs, and separately approved cleanup execute SQL. The six other destructive patterns per tool must all be denied. A missing form may mean SQL executed unexpectedly: stop writes immediately, report the failure, and use read-only diagnostics if needed. Also stop on an unexpected result, a changed target, changed state after Deny/Cancel, or a repeated form after an accepted call. Check the catalog before reading a table that may have been dropped. Do not retry, rewrite SQL, change a migration name to bypass a failure, or continue to cleanup after a stop.

1. Create two small fixtures

Choose one identifier-safe, unique schema name mcp_elicitation_test_<UTC>_<random>. Replace <schema> below with that exact name. Use a read-only catalog check to confirm it does not exist; never reuse an existing schema. Keep a record of every object and migration created by this run. Fully qualify all fixture references. Do not touch existing schemas, other objects, roles, or extensions.

Before every write call, show the exact target, tool, SQL, and migration name if applicable. Run calls sequentially, never concurrently.

Through execute_sql, run:

CREATE SCHEMA <schema>;
CREATE TABLE <schema>.sql_rows (
  id int PRIMARY KEY,
  n int NOT NULL,
  keep int,
  doomed int
);
INSERT INTO <schema>.sql_rows (id, n, keep, doomed)
VALUES (1, 10, 100, 1000), (2, 20, 200, 2000), (3, 30, 300, 3000);

Through apply_migration, use the unique name <schema>_setup and run:

CREATE TABLE <schema>.migration_rows (
  id int PRIMARY KEY,
  n int NOT NULL,
  keep int,
  doomed int
);
INSERT INTO <schema>.migration_rows (id, n, keep, doomed)
VALUES (1, 10, 100, 1000), (2, 20, 200, 2000), (3, 30, 300, 3000);

Both setup calls are benign baselines: expect no server elicitation form. Also check the bounded UPDATE baseline once per tool, expecting no form:

execute_sql:
UPDATE <schema>.sql_rows SET n = n WHERE id = 1;

apply_migration, name <schema>_bounded_update:
UPDATE <schema>.migration_rows SET n = n WHERE id = 1;

If any benign baseline unexpectedly prompts, stop and report it; do not accept it.

Use read-only execute_sql calls to save an initial snapshot of each table's existence, column definitions, and rows ordered by id. Both must contain exactly the three seeded rows. Use this row-query template with the appropriate table:

SELECT id, n, keep, doomed FROM <schema>.<table> ORDER BY id;

Keep these same two tables throughout. Do not reseed or create replacement fixtures between cases.

2. Check response handling once per tool

Complete the whole execute_sql sequence first, then the whole apply_migration sequence. Each uses its own table:

execute_sql:
UPDATE <schema>.sql_rows SET n = n + 1;

apply_migration, name <schema>_response_update:
UPDATE <schema>.migration_rows SET n = n + 1;

For each tool, submit that exact SQL in this order. For apply_migration, keep the same name and SQL across all attempts:

- Deny/Decline: expect a server form and wait for me to deny it. Read back the fixture. Its schema and all rows must match the initial snapshot, including n = 10, 20, 30.
- Cancel: only if the client exposes a distinct cancellation of the server form, submit the same call and wait for me to cancel it. Read back the fixture and require the same unchanged snapshot. If no distinct cancellation is available, record NOT OBSERVABLE and skip this attempt. A CLI interrupt or another Deny is not cancellation evidence.
- Accept: submit the same call and wait for me to accept the form once. Read back the fixture. Expect exactly n = 11, 21, 31 by id, with all other data and column definitions unchanged. Save this as the post-accept snapshot. Do not retry after success.

Only these two test UPDATEs are accepted. Each affects the three rows in its own fixture. Their readbacks check this run's outcome, not a general exactly-once guarantee.

3. Check the remaining SQL patterns with Deny only

Run the following six cases once through execute_sql on sql_rows, then once through apply_migration on migration_rows. Replace <table> accordingly. Every case must produce a server elicitation form; wait for me to choose Deny/Decline every time.

Case                   SQL                                                        apply_migration name
D1 delete with WHERE   DELETE FROM <schema>.<table> WHERE id = 1;                  <schema>_d1
D2 delete all          DELETE FROM <schema>.<table>;                               <schema>_d2
D3 truncate            TRUNCATE TABLE <schema>.<table>;                            <schema>_d3
D4 drop table          DROP TABLE <schema>.<table>;                                <schema>_d4
D5 drop column         ALTER TABLE <schema>.<table> DROP COLUMN doomed;            <schema>_d5
D6 drop shorthand      ALTER TABLE <schema>.<table> DROP doomed;                   <schema>_d6

Do not repeat the unbounded UPDATE case, which phase 2 already covers. Do not accept or cancel any detection case. Do not reset fixtures or read them back after individual detection cases. The tables and doomed columns must remain available because every case is denied. No migration-history write is expected for these denied apply_migration calls.

At the end of each tool's six-case block, make one read-only comparison of the table's existence, column definitions, and ordered rows against its post-accept snapshot. Expect n = 11, 21, 31 and no other changes. This is an aggregate check for the block, not a per-case readback. If a form is missing or another stop condition occurs mid-block, stop immediately rather than finishing the block.

4. Report before cleanup

Return a compact table with tool, case/SQL reference, expected form, observed form, manual action, result, and PASS/FAIL/NOT OBSERVABLE. Include the benign baselines, response checks, and each detection case. Distinguish tool results from my observations of the actual forms. Report the initial, response, and aggregate readback comparisons, any skipped cancellation, and any stop condition. List the schema and migration names created by this run.

With cancellation available, the response and detection phases contain 18 form actions: 2 Accept, 14 Deny, and 2 Cancel. Without it, they contain 16 actions and two cancellation checks marked NOT OBSERVABLE. Setup, benign baselines, read-only checks, and cleanup are outside those counts.

Keep the conclusion limited to manual form behavior and observed database outcomes. This does not prove SDK authentication, request binding, replay protection, or that classification was skipped on the accepted retry. That last claim needs the separate automated issued-state test whose classifier throws if called. Comment, CTE, and DO-block regex limitations are outside this run; do not add experiments or runnable examples for them. Do not add resource, performance, deep-nesting, or parser cases.

5. Cleanup only after separate approval

After the report, ask whether I want cleanup. If approved, use read-only checks to list the schema's contents and confirm every contained object belongs to this run. If ownership is uncertain, stop and leave it intact.

Show the exact target and SQL, then call execute_sql with:

DROP SCHEMA <schema> CASCADE;

Expect a server elicitation form and wait for my manual Accept. If I decline or cancel, leave the fixtures in place. If the form is missing, stop and report the failure, using only read-only diagnostics if needed. After an accepted cleanup, verify schema absence with a read-only catalog check. Never delete migration history, modify client configuration, or stop the server.
```
</details>

## Trade-offs

- Best-effort regex is not a security barrier: literal-comment, UPDATE-comment and WITH/DO misses remain.
- Clients without elicitation support maintain current behaviour: SQL can execute unprompted. 
- Hosted-platform SQL opt-in is separate ([#38211](https://github.com/supabase/platform/pull/38211)), defaults to off behind a feature flag.


Rodriguespn (changes requested):
Asking for changes as I identified some edge cases on the parsing that I think would need further discussion. 

For example, a two-line SQL statement that starts with a comment (`--`) isn't sanitized correctly today:

```sql
SELECT '--';
DROP TABLE films;
```

I tried this locally against the current regex and the table gets dropped with no confirmation prompt. The `--` inside the string literal on line 1 is treated as a real comment and silently eats the rest of that line, including the `;` that would otherwise separate it from the DROP.

Any reason why we're doing our own parsing instead of using the official Postgres parser  [`libpg-query-node`](https://github.com/constructive-io/libpg-query-node)? 

We're already using it on the backend to classify `execute_sql` queries: [`sql-classification.worker.ts`](https://github.com/supabase/platform/blob/650d7a708c6920c879323862009d8588460659b3/api/apps/mgmt-api/src/lib/mcp/sql-classification.worker.ts).

Same scenario through the real parser:

```ts
import { parse } from '@libpg-query/parser'

const { stmts } = await parse("SELECT '--';\nDROP TABLE films;")
// stmts.length === 2
// stmts[1].stmt -> { DropStmt: { removeType: 'OBJECT_TABLE', objects: [...films] } }
```

Two real statements come out and the DROP is fully visible, no regex needed, because the parser understands string literals natively.

I found a few other cases the current regex misses the same way: 
1. `UPDATE` without a `WHERE` when a comment sits nearby (leading, inline, or trailing), 
2. `ALTER TABLE t DROP col` without the optional `COLUMN` keyword,
3. destructive statements wrapped in `WITH ...` or `DO $$ ... $$`. 

All of these come for free with a real parser (checking the `whereClause` field, matching on the `AlterTableCmd` subtype instead of literal text, or just walking the statement list) instead of patching the regex edge case by edge case.

I've prepared some [test cases](https://linear.app/supabase/issue/AI-983/mcp-investigate-adding-observability-to-execute-sql-tool#test-cases-6391c182) that we can reuse here when I was working on the sql parsing.

Lmk what you think.

Rodriguespn (changes requested):
in progress

barryroodt:
- 20 mandatory manual cases passed, 10 each for `execute_sql` and `apply_migration`. 
- `UPDATE` without `WHERE` verified genuine decline, cancel and accept flows for both tools. 
- All expected no-prompt cases passed.

Three optional `execute_sql` cases executed without a confirmation form:

- `SELECT '--'; DROP TABLE <schema>.<table>;` dropped the table.
- `WITH changed AS (DELETE FROM <schema>.<table> WHERE id=1 RETURNING id) SELECT * FROM changed;` deleted the row.
- `DO $$ BEGIN DELETE FROM <schema>.<table> WHERE id=1; END $$;` deleted the row.

These are known regex coverage limits, not new regressions. Follow-up PRs introducing a SQL-to-AST parser are inbound to address these gaps.

gregnr (approved):

Rodriguespn (approved):
Approving to unlock this

PR #409, opened 2026-09-03
chore: handle schemas with dot separator

staaldraad:

chore / fix

Schemas are expected to never contain a `.` (dot)

## What is the new behavior?

Schemas may contain a `.` (dot) and will be treated the same as a schema without.

Extends #407


barryroodt (approved):
Looks good! thanks!

PR #410, opened 2026-09-03
fix: discourage host-filesystem access in raw-SQL tool descriptions

stoallan:
## What

Appends a shared `HOST_BOUNDARY_NOTE` to the descriptions of the two raw-SQL tools — **`execute_sql`** and **`apply_migration`** — and a matching one-line reminder in the server `instructions`, discouraging the model from reading server-side files or running OS commands via SQL (`COPY … FROM '/path'`, `COPY … FROM PROGRAM`, `pg_read_file`, `pg_ls_dir`, `lo_import`).

## Why

Cut **tripwire noise on the managed fleet.** LLMs driving the MCP attempt host-filesystem reads via SQL, and the tripwire fires on the **attempt** — it records that the SQL ran, not whether it succeeded — so every probe generates alert noise regardless of outcome. Discouraging the model from making the attempt reduces the volume at the source. This is the exfil-OUT counterpart to the injection-IN guidance already present (AI-794 / `wrapWithUntrustedDataBoundary`).

## Design

- **On the tool descriptions, not only `server` instructions.** Per the 2026 MCP spec, `instructions` rides on the *optional* `server/discover` response and isn't guaranteed to be injected; a client cannot invoke a tool without ingesting its description — so the description is the reliable delivery vehicle. The server-instructions line is added too, as opportunistic defense-in-depth.
- **Both raw-SQL tools.** `apply_migration` forwards arbitrary SQL too; these two are the only raw-SQL entry points.
- **A shared constant** (`HOST_BOUNDARY_NOTE`) to keep the two descriptions from drifting, mirroring the existing `wrapWithUntrustedDataBoundary` factoring.
- **Deployment-neutral wording.** The text names no deployment context — it's true everywhere. This matters because the same `mcp-server-supabase` also ships to self-hosted Studio, where a "Supabase-managed infrastructure" rationale would be false (the operator owns the host). No self-hosted benefit is claimed; the goal is purely managed noise.

## FYI (per review)

ChatGPT snapshots/freezes MCP tool descriptions + instructions, so this needs a plugin resubmission to take effect there — tracked on the Supabase side, no action here.

## Testing

String-concat append to two tool descriptions + one instructions bullet (type-trivial). No test asserts the raw-SQL description text (the `untrusted user data` assertions target the result wrapper), and the instructions test compares against the exported constant, so it stays in sync. Local suite not run — CI validates format/typecheck/tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -94,6 +94,7 @@ Here are guidelines for using Supabase tools effectively:
 - Before making schema changes, inspect the existing tables so you understand the current structure
 - When debugging issues, start by reading the project's logs and its security and performance advisories before making changes
 - Look up the project's API URL and its publishable API keys when helping users configure client-side integrations
+- The project database runs on shared Supabase infrastructure; its host filesystem and OS are not the user's. Do not read server-side files or run OS commands via SQL.
 
 If you have access to a local development environment with a filesystem and shell:
 - Install the Supabase agent skill for critical development and security guidance: \`npx skills add supabase/agent-skills\` (https://supabase.com/docs/guides/getting-started/ai-skills.md)
--- a/packages/mcp-server-supabase/src/server.ts
+++ b/packages/mcp-server-supabase/src/server.ts
@@ -94,7 +94,7 @@ Here are guidelines for using Supabase tools effectively:
 - Before making schema changes, inspect the existing tables so you understand the current structure
 - When debugging issues, start by reading the project's logs and its security and performance advisories before making changes
 - Look up the project's API URL and its publishable API keys when helping users configure client-side integrations
-- The project database runs on shared Supabase infrastructure; its host filesystem and OS are not the user's. Do not read server-side files or run OS commands via SQL.
+- SQL runs on the project's database server. Never read server-side files or run OS commands via SQL.
 
 If you have access to a local development environment with a filesystem and shell:
 - Install the Supabase agent skill for critical development and security guidance: \`npx skills add supabase/agent-skills\` (https://supabase.com/docs/guides/getting-started/ai-skills.md)
--- a/packages/mcp-server-supabase/src/tools/database-operation-tools.ts
+++ b/packages/mcp-server-supabase/src/tools/database-operation-tools.ts
@@ -112,6 +112,13 @@ const executeSqlOutputSchema = z.object({
   result: z.string(),
 });
 
+// Appended to both raw-SQL tool descriptions (execute_sql, apply_migration),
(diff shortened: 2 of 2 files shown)


Rodriguespn (changes requested):
Hey Allan, this looks good — clean approach, and I've verified the note lands on exactly `execute_sql` + `apply_migration` and the instructions line ships on initialize. One thing I'd like before we merge:

The clause ("shared Supabase infrastructure; its host filesystem and OS are not the user's") is accurate for hosted/mgmt-api, where the alerts came from, but the same string ships to self-hosted Studio ([here](https://github.com/supabase/supabase/blob/master/apps/studio/lib/api/self-hosted/mcp.ts)), where the user owns the host, and there it's actually load-bearing, since installs predating the superuser removal connect as `supabase_admin`, where these primitives genuinely succeed. 

So I'd make the wording deployment-neutral so it's true everywhere:

```ts
const HOST_BOUNDARY_NOTE =
  " Never read server-side files or run OS commands via SQL (e.g. `COPY ... FROM '/path'`, `COPY ... FROM PROGRAM`, `pg_read_file`, `pg_ls_dir`, `lo_import`) — load external data from the client side instead.";
```

And in `server.ts`:
```diff
-- The project database runs on shared Supabase infrastructure; its host filesystem and OS are not the user's. Do not read server-side files or run OS commands via SQL.
++ SQL runs on the project's database server. Never read server-side files or run OS commands via SQL.
```

Instruction and examples unchanged, only the rationale. Keeps the single shared constant, no per-platform branch, no drift.

Two small nits, non-blocking: 
1. qualify the "role isn't a superuser" line in the PR body (true for hosted, not for old self-hosted installs), 
2. Note in the PR that ChatGPT snapshots/freezes tool descriptions + instructions, so it needs a plugin resubmission to take effect there. No action from your side needed, this one is on us. Also, low risk since the MCP connection on Chatgpt is serverless and (hopefully) sandboxed by OpenAI

Rodriguespn (approved):
Round 2: verified. HOST_BOUNDARY_NOTE is a single shared constant appended to both execute_sql and apply_migration descriptions, deployment-neutral (no "shared Supabase infrastructure"/"not the user's" language), keeps the imperative and all four examples (COPY ... FROM '/path', COPY ... FROM PROGRAM, pg_read_file, pg_ls_dir, lo_import). server.ts guideline line is now deployment-neutral: "SQL runs on the project's database server. Never read server-side files or run OS commands via SQL." Only the two expected files touched, no per-platform branching, no wording drift between the two tool descriptions. Addresses my round-1 feedback. Approving.

PR #411, opened 2026-09-03
feat(mcp): hide legacy cost tools from form-capable clients

mattrossman:
Makes `tools/list` adapt to `costConfirmation.enabledTools` to prevent supported clients from getting prompted twice, from both the legacy tools and elicitation strategy, surfaced when testing https://github.com/supabase/platform/pull/37717

The `tools` callback now receives the per-request `ServerContext` so `tools/list` can vary by client: 
- If both `create_project` and `create_branch` are in `costConfirmation.enabledTools`, `get_cost` and `confirm_cost` are no longer advertised. They stay callable, so a client that already knows the names keeps working.
- If only one `create_` tool is enabled, the two cost tools are still advertised because the other `create_` tool still needs a `confirm_cost_id`, but their `type` argument is narrowed.
- `create_project` and `create_branch` drop `confirm_cost_id` when they're in `enabledTools`, since a form-capable client confirms inline and never needs it.

Legacy clients are unaffected.

## How to review

Run the server over HTTP w/ `costConfirmation` enabled for both tools and connect a form-capable client (e.g. Claude Code w/ the SDK v2 flag as described [here](https://github.com/supabase/mcp/pull/394#pullrequestreview-5097736337)). The tool list should have no `get_cost` or `confirm_cost`, and `create_project` no longer has a `confirm_cost_id` argument. Connect again without declaring `elicitation` and both tools are back.

## Notes

I renamed the test helper `setupFormCapable` to `setupModern` and made its capabilities explicit via a `FORM_CAPABLE` constant, because technically legacy clients can have form capabilities.

The #391 test "a supplied `confirm_cost_id` cannot bypass the form" is now "is rejected". The field isn't in a form-capable client's schema anymore, so strict arg parsing errors instead of routing to the form.

### DIFF
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -682,6 +682,36 @@ describe('tools', () => {
       expect(names).toContain('create_project');
     });
 
+    test('omits confirm_cost_id from create_project for a form-capable client', async () => {
+      const { client } = await setupModern({
+        clientCapabilities: FORM_CAPABLE,
+      });
+
+      const { tools } = await client.listTools();
+      const createProjectTool = tools.find(
+        (tool) => tool.name === 'create_project'
+      );
+
+      expect(createProjectTool?.inputSchema.properties).not.toHaveProperty(
+        'confirm_cost_id'
+      );
+    });
+
+    test('omits confirm_cost_id from create_branch for a form-capable client', async () => {
+      const { client } = await setupModern({
+        clientCapabilities: FORM_CAPABLE,
+      });
+
+      const { tools } = await client.listTools();
+      const createBranchTool = tools.find(
+        (tool) => tool.name === 'create_branch'
+      );
+
+      expect(createBranchTool?.inputSchema.properties).not.toHaveProperty(
+        'confirm_cost_id'
+      );
+    });
+
     test('narrows cost tools to branch while create_branch still needs confirm_cost_id', async () => {
       const { client } = await setupModern({
         clientCapabilities: FORM_CAPABLE,
@@ -4161,7 +4191,7 @@ describe('tools', () => {
       }
     });
 
-    test('form-capable client: a supplied confirm_cost_id cannot bypass the form', async () => {
+    test('form-capable client: a supplied confirm_cost_id is rejected', async () => {
       const { client } = await setupModern({
         clientCapabilities: FORM_CAPABLE,
       });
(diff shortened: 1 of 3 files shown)


packages/mcp-server-supabase/src/server.ts
```diff
@@ -253,9 +257,52 @@ export function createSupabaseMcpServer(options: SupabaseMcpServerOptions) {
...
+      // drop `confirm_cost_id` and the legacy cost tools only offer the types
+      // that still need them. With nothing left to quote they are hidden
+      // entirely.
+      if (
+        costConfirmationCodec &&
+        costConfirmation &&
+        ctx &&
+        isFormCapable(ctx)
+      ) {
+        for (const name of costConfirmation.enabledTools) {
```
barryroodt: Non-blocking, but would it make sense to scope this loop to the two create tools rather than everything in `costConfirmation.enabledTools`? That way the block stays cost-specific as `enabledTools` grows.

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -167,21 +172,29 @@ const COST_CONFIRMATION: NonNullable<
...
- * era, so the form-capable lane - which depends on the per-request `_meta`
+ * elicitation lanes: a client pinned to the 2026-07-28 protocol. Raw
+ * `StreamTransport` only speaks the 2025 era, so the form-capable lane - which
+ * depends on the per-request `_meta`
  * envelope - needs the same in-process HTTP transport the hosted runtime
  * uses.
  */
-async function setupFormCapable(options: FormCapableSetupOptions = {}) {
-  const { readOnly, projectId, elicitationAction } = options;
+async function setupModern(options: ModernSetupOptions = {}) {
```
barryroodt, line 190: Happy with the `setupModern` rename, non-blocking. Heads-up that #408 touches this same region and will conflict, I'll sort that out on the #408 side.

packages/mcp-server-supabase/src/server.ts
mattrossman: Tweaked to be explicit about the two cost tools in https://github.com/supabase/mcp/pull/411/changes/f53073a6fabefe1306504d4fdaa86cf5d4a4c011

packages/mcp-server-supabase/src/server.test.ts
mattrossman: Thanks for the heads up!

barryroodt (approved):
Looks good. Thanks for scoping the omit loop, reads clearly now.

PR #412, opened 2026-09-03
feat: URL-mode secret collection for create_edge_function_secret

barryroodt:
## Why

Add an Edge Function’s third-party API key without pasting it into the conversation.

## What changed

- Secret values are entered only in Dashboard, never the AI client/model. Local HTTP enables this by default.
- Discovery requires modern per-request URL capability; existing functions, read-only, auth and platform-support gates remain.
- Breaking: move confirmation signing to `elicitation.requestState:{key,principal,ttlSeconds?}` (`requestStateKey` becomes `key`).
- Optional `elicitation.confirmation:{enabledTools:readonly ElicitationToolName[]}` selects unified cost/SQL forms; `enabledTools` is explicit, with no implicit library all-tools default, and `[]` disables forms.
- `elicitation.secretCollection:{connectUrlTemplate}` is independently optional.

## How to test

**Optional manual checks—not reported as run.** Use existing credentials and authorized test resources. Accepting cost/SQL actions or saving a secret makes real writes and may incur costs.

Follow [setup](https://github.com/supabase/mcp/blob/bdb744f21339e81a73f0cfc3c1231d8244d01c33/CONTRIBUTING.md#development-setup), install dependencies, then run `pnpm dev:http --port 3111 --api-url https://api.supabase.green`.

Connect Inspector 2.5.0 Modern with both form and URL support to `http://127.0.0.1:3111/mcp?project_ref=REF&read_only=false` with `Authorization: Bearer <PAT>`; omit `features` and `skip_elicitations` for normal local defaults.

- **Cost:** remove `project_ref` for an unscoped connection with account tools available. On an eligible billable `create_project` path, inspect and decline the confirmation to avoid creating a project. Restore `project_ref=REF` for the SQL and secret checks.
- **SQL:** use disposable project/data you are authorized to modify. Inspect and decline destructive `execute_sql`/`apply_migration` confirmations.
- **Secret:** with Dashboard authentication/write permission and the matching deployed route, call `create_edge_function_secret` with a fresh `name` (no value). Save only in Dashboard, return to Inspector and accept within 120s; expect `stored:true`. Legacy/non-URL clients should not list the tool.

## Trade-offs

- Local HTTP signed state lasts 120s; omitted library TTL remains 600s.
- Separate timestamp-based fresh-call recovery checks for updates within 600s; it does not extend signed-state validity or describe URL validity.
- Original-request acceptance checks `updated_at>=issued_at`; `stored:true` is not a receipt; another collaborator’s write can satisfy it.
- Recovery can return recent success; a fresh name avoids that shortcut.
- Custom API origins require `--secret-url-template`.

### DIFF
--- a/packages/mcp-server-supabase/src/server.test.ts
+++ b/packages/mcp-server-supabase/src/server.test.ts
@@ -28,6 +28,7 @@ import {
   mockBranches,
   mockContentApiSchemaLoadCount,
   mockProjects,
+  mockSecrets,
   setupMockApis,
 } from '../test/mocks.js';
 import { createSupabaseApiPlatform } from './platform/api-platform.js';
@@ -212,6 +213,61 @@ async function setupFormCapable(options: FormCapableSetupOptions = {}) {
     { name: MCP_CLIENT_NAME, version: MCP_CLIENT_VERSION },
     {
       capabilities: { elicitation: { form: {} } },
+      versionNegotiation: { mode: { pin: MODERN_PROTOCOL_VERSION } },
+      ...(elicitationAction === undefined && {
+        inputRequired: { autoFulfill: false },
+      }),
+    }
+  );
+
+  if (elicitationAction !== undefined) {
+    client.setRequestHandler('elicitation/create', async () =>
+      elicitationAction === 'accept'
+        ? { action: 'accept' as const, content: {} }
+        : { action: elicitationAction }
+    );
+  }
+
+  await client.connect(transport);
+
+  return { client };
+}
+
+/**
+ * Sets up an MCP client with URL elicitation capability for the
+ * `create_edge_function_secret` secret-collection elicitation lane.
+ */
+async function setupUrlCapable(options: FormCapableSetupOptions = {}) {
+  const { readOnly, projectId, elicitationAction } = options;
+
+  const platform = createSupabaseApiPlatform({
+    accessToken: ACCESS_TOKEN,
+    apiUrl: API_URL,
+  });
+
+  await platform.init?.({
+    clientInfo: { name: MCP_CLIENT_NAME, version: MCP_CLIENT_VERSION },
+    clientCapabilities: { elicitation: { url: {} } },
+  });
+
+  const handler = createSupabaseMcpHandler({
+    platform,
+    projectId,
+    readOnly,
+    costConfirmation: COST_CONFIRMATION,
+    secretCollection: {
+      connectBaseUrl: 'https://supabase.com/dashboard/mcp_callback',
+    },
+  });
+
+  const transport = new StreamableHTTPClientTransport(MCP_ENDPOINT, {
+    fetch: (url, init) => handler.fetch(new Request(url, init)),
+  });
+
+  const client = new Client(
+    { name: MCP_CLIENT_NAME, version: MCP_CLIENT_VERSION },
+    {
+      capabilities: { elicitation: { url: {} } },
       versionNegotiation: { mode: { pin: MODERN_PROTOCOL_VERSION } },
       ...(elicitationAction === undefined && {
         inputRequired: { autoFulfill: false },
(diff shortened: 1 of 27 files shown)


packages/mcp-server-supabase/src/tools/cost-confirmation.ts
```diff
@@ -35,7 +35,23 @@ export type BranchCostState = {
  * Signed `requestState` payload for any cost-confirmation elicitation this
  * server issues, discriminated by `tool`.
  */
-export type CostConfirmationState = ProjectCostState | BranchCostState;
+export type CostConfirmationState =
```
mattrossman, line 38: Should this be called `CostConfirmationState` still, if it's for more than costs?

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -4475,6 +4526,733 @@ describe('tools', () => {
...
+        expect(result.isError).toBe(true);
+        const textContent = result.content.find((c: any) => c.type === 'text');
+        expect((textContent as any)?.text).toContain(
+          'This client cannot open a browser page'
+        );
+        expect(JSON.stringify(result)).not.toContain('http');
+      }
+    });
+
+    test('tool input schema exposes project_id, name and replace, never value', async () => {
```
mattrossman, line 4635: Nit, seems like this could be tested more clearly w/ a single assertion of the strict list of properties, e.g. `expect(Object.keys(secretTool?.inputSchema.properties ?? {}).sort()).toEqual(['name', 'project_id', 'replace']);`

Asserting that there isn't a "value" property doesn't add much if someone used a different property name like "secret".

packages/mcp-server-supabase/src/server.test.ts
```diff
@@ -4475,6 +4526,733 @@ describe('tools', () => {
...
+      const project = await createProject({
+        name: 'Project 1',
+        region: 'us-east-1',
+        organization_id: org.id,
+      });
+      project.status = 'ACTIVE_HEALTHY';
+
+      const first = (await client.request(
+        {
+          method: 'tools/call',
```
mattrossman, line 4692: Should we be changing the `callTool()` helper to shorten these tests? Seems like we could use that halper to avoid the raw JSON-RPC stuff (same for cost confirmation tests)

mattrossman: We're due to condense this file overall and that can be a follow up thing, I'm just trying to find ways to make it easier to review as 1000+ LOC of tests in the PR is a little hard to meaningfully vet.

packages/mcp-server-supabase/src/tools/secret-tools.ts
```diff
@@ -0,0 +1,239 @@
...
+    .max(256)
+    .regex(/\S/)
+    .refine((n) => !n.startsWith('SUPABASE_'), {
+      message: 'Secret names starting with SUPABASE_ are reserved.',
+    }),
+  replace: z
+    .boolean()
+    .optional()
+    .describe(
+      'Set to true to ask the user for a new value even if this secret was updated in the last 10 minutes. Default false: a recent update is reported as stored without asking again.'
```
mattrossman, line 49: What's the rationale for the 10 minute cutoff here? IIUC this means if we run the tool with `replace: false`, it can still replace an existing value that's older than 10 minutes. I'd expect this flag to behave like a regular upsert flag, where `replace: false` means I only write the value if it wasn't set.

packages/mcp-server-supabase/src/transports/local-http-entry.test.ts
```diff
@@ -141,12 +221,209 @@ describe('startLocalHttpEntry', () => {
...
+
+  test('does not replace an explicitly empty template with a default', async () => {
+    await entry.close();
+    entry = await startEntry({ secretUrlTemplate: '' });
+    const response = await fetch(entry.url, {
+      method: 'POST',
+      headers: { ...AUTH_HEADERS, 'content-type': 'application/json' },
+      body: JSON.stringify({ jsonrpc: '2.0', id: 1, method: 'tools/list' }),
+    });
+    expect(response.status).toBe(500);
```
mattrossman, line 312: Nit, should we be rejecting invalid template URL on server start instead of request time?

packages/mcp-server-supabase/src/tools/secret-tools.ts
```diff
@@ -0,0 +1,239 @@
...
+            isError: true,
+          };
+        }
+
+        if (state.project_id !== project_id || state.name !== name) {
+          return {
+            content: [
+              {
+                type: 'text' as const,
+                text: 'Request state arguments do not match the current arguments.',
```
mattrossman, line 182: Could use test coverage here

packages/mcp-server-supabase/src/transports/local-http-entry.ts
```diff
@@ -56,25 +59,53 @@ export function describeRequest(body: unknown): string {
...
     : 'unknown';
-  return [
+  const description = [
     (method ?? 'unknown').padEnd(28),
     name.padEnd(24),
     `(${protocolVersion ?? 'legacy'})`,
   ].join('  ');
+  return elicitation === undefined
+    ? description
+    : `${description}  elicitation=${elicitation}`;
```
mattrossman, line 81: Is this valuable to have on each line? Wondering if that's a leftover debugging artifact, or maybe it could be filtered to only show on tools that use elicitations. Otherwise seems like dev noise for most tools.

mattrossman (approved):
I tested this in the MCP Inspector since AFACT no current coding agents implement URL mode elicitations w/ modern protocol properly, and it generally worked as expected 👍 

As long as it's behind a flag for us to test first I'm okay to proceed, comments above are generally non blocking but some are worth discussing.

packages/mcp-server-supabase/src/tools/secret-tools.ts
barryroodt: The ten-minute window lets a fresh or retried call pick up a recent Dashboard save without asking for the secret again. `replace: true`  forces fresh entry; `false` allows that recovery shortcut rather than guaranteeing insert-only behaviour, so an older secret can still  prompt again.

PR #419, opened 2026-09-08
feat(management): add v2 api client

jordienr:
## Problem

Health Advisor is exposed through Management API v2, but the MCP only has a generated v1 client.

## Fix

Generate and validate v2 Management API types alongside v1, and add a typed v2 client with the existing bearer-token behavior.

## How to test

- Run 
> @supabase/mcp-server-supabase@0.12.0 generate:management-api-types /Users/jordi/Documents/supabase/projects/supabase/mcp/packages/mcp-server-supabase
> pnpm generate:management-api-v1-types && pnpm generate:management-api-v2-types

> @supabase/mcp-server-supabase@0.12.0 generate:management-api-v1-types /Users/jordi/Documents/supabase/projects/supabase/mcp/packages/mcp-server-supabase
> node scripts/generate-management-api-types.mjs

> @supabase/mcp-server-supabase@0.12.0 generate:management-api-v2-types /Users/jordi/Documents/supabase/projects/supabase/mcp/packages/mcp-server-supabase
> node scripts/generate-management-api-v2-types.mjs
- Expected result: both v1 and v2 type files regenerate without a diff.
- Run 
> @supabase/mcp-server-supabase@0.12.0 typecheck /Users/jordi/Documents/supabase/projects/supabase/mcp/packages/mcp-server-supabase
> tsc --noEmit
- Expected result: type checking passes.


packages/mcp-server-supabase/scripts/generate-management-api-v2-types.mjs
```diff
@@ -0,0 +1,26 @@
...
+);
+
+function transform(schemaObject, options) {
+  if (options.path?.endsWith('JsonValue_Output')) {
+    return UNKNOWN;
+  }
+  return undefined;
+}
+
+const ast = await openapiTS(SPEC_URL, { transform });
```
barryroodt: This duplicates the existing generator, including the recursive-JSON workaround, with only the spec URL and output path changing. Could we keep one generator with explicit v1/v2 URL and output-path pairs, so future fixes only need to happen in one place?

barryroodt (approved):
Thanks, looks good with one non-blocking request. Could we consolidate the two type generators into one script so the shared generation logic lives in one place? Details inline.

packages/mcp-server-supabase/scripts/generate-management-api-v2-types.mjs
jordienr: [Codex] fixed in ac90855

PR #444, opened 2026-09-22
Joshenlim/ai 1240 mcp add functionality for listing and getting notebooks

joshenlim:
## Context

_Prefacing this that it's my first time contributing to this repo, so I've had substantial help from Claude for this 😅 🙏_
Please do let me know if there's anything that i can do better RE the changes here! 🙂 🙏 

This PR adds 2 new MCP tools for managing the new notebooks for projects - `list_notebooks` and `get_notebook`, which pings the newly added `v2` management API endpoints for notebooks ([ref](https://github.com/supabase/platform/pull/38167))

## To test

This is what I've done to verify that the MCP tooling is working correctly - let me know if there's a better way to do this! 🙏  I seem to be running into permission issues on the API end however, hence why the PR is still in draft at the moment as I'm unable to test E2E properly.

- Am using the `@modelcontextprotocol/inspector` GUI to test the MCP server locally (via `pnpm dev:http`)
- Connected to the MCP server via `http://127.0.0.1:3111/mcp?project_ref=dfbjnxjedvxmmeuiuygg&features=notebooks`
  - Using a production project ref that has some notebooks
- Able to connect successfully as such:
  image
- And able to view the new tools created:
  image
- However trying to run `list_notebooks` throws me this API error:
  image


packages/mcp-server-supabase/src/platform/api-platform.ts
```diff
@@ -840,20 +850,67 @@ export function createSupabaseApiPlatform(
     },
   };
 
+  const notebooks: NotebookOperations = {
+    async listNotebooks(projectId: string) {
```
barryroodt, line 854: `/v2/projects/{ref}/notebooks` is cursor-paginated, but this makes one request and drops `links`. On a project with more notebooks than the default page size, the list silently stops at page one, and the model will assume the rest don't exist. The mock in `test/mocks.ts` always returns `links.next: null`, so the tests can't catch this. Could you either follow `links.next` / `page[after]` until it runs out, or expose the cursor (and maybe `filter[name]`) as tool params and return the next cursor? Please also add a test with more than one page of notebooks.

packages/mcp-server-supabase/src/tools/notebook-tools.ts
```diff
@@ -0,0 +1,80 @@
...
+      title: 'List notebooks',
+      readOnlyHint: true,
+      destructiveHint: false,
+      idempotentHint: true,
+      openWorldHint: false,
+    },
+  },
+  get_notebook: {
+    description:
+      'Gets a notebook from a Supabase project, including its cells.',
```
barryroodt, line 45: Notebook cells are markdown and SQL written by anyone with project access, and get_notebook returns them raw. A malicious cell could steer an agent that also has execute_sql. execute_sql already guards against this: its description warns about untrusted data, and results are wrapped in `<untrusted-data-uuid>` boundaries (`src/tools/util.ts:94`). Could you add the same warning here and wrap cell content the same way?

barryroodt (changes requested):
Thanks for this, @joshenlim. Two small changes before merge, both inline: pagination in listNotebooks, and marking get_notebook output as untrusted.

packages/mcp-server-supabase/src/tools/notebook-tools.ts
joshenlim: addressed [here](https://github.com/supabase/mcp/pull/444/commits/120962512eab14cc3b55be00124ac74097c361b7)!

packages/mcp-server-supabase/src/platform/api-platform.ts
joshenlim: aaand addressed [here](https://github.com/supabase/mcp/pull/444/commits/50b1c5753daf798dbcbad07b6b1cf9a36c9c411f)!

barryroodt (approved):

